PR #29, opened 2025-09-27
Fix three onboarding issues

aaronjeline:
Fixed three issues I ran into while onboarding

1. The build-app script assumes you have access to Niko's singing material
2. The "open project" button returns you to the main screen if you try to open a directory not initialized as a Symposium project. It now displays an error message before doing this. (I had initially assumed I _added_ symposium to an existing project)
3. The bare git clone did not set up the git references for the remote, causing taskspace creation to fail. 

Let me know if you want any stylistic changes.

### DIFF
--- a/.kiro/specs/project-selection-error-handling/design.md
+++ b/.kiro/specs/project-selection-error-handling/design.md
@@ -1,0 +1,242 @@
+# Design Document
+
+## Overview
+
+This design addresses the user experience issue where the macOS Symposium app silently fails when users select invalid directories during project selection. The solution introduces comprehensive error handling with informative user feedback while maintaining the existing validation logic and app flow.
+
+## Architecture
+
+### Current Flow Analysis
+
+The current project selection flow follows this path:
+1. User clicks "Open Existing Project" → `ProjectSelectionView`
+2. File picker opens → `fileImporter` modifier
+3. User selects directory → `openProject(at: url.path)`
+4. Calls `onProjectSelected(path)` → `handleProjectSelected` in `App.swift`
+5. Sets `settingsManager.activeProjectPath = path`
+6. Triggers `appStart()` → `runStartupLogic()`
+7. Validates using `Project.isValidProjectDirectory(path)`
+8. If invalid: clears path and shows project selection (silent failure)
+
+### Proposed Architecture Changes
+
+The design introduces error handling at two key points:
+
+1. **Immediate Validation**: Add validation in `ProjectSelectionView` before calling `onProjectSelected`
+2. **Fallback Validation**: Enhance existing validation in `runStartupLogic()` with user feedback
+
+This dual approach ensures users get immediate feedback while maintaining backward compatibility with the existing app state machine.
+
+## Components and Interfaces
+
+### 1. Error Types
+
+Create a new `ProjectValidationError` enum to represent different validation failure scenarios:
+
+```swift
+enum ProjectValidationError: LocalizedError {
+    case missingProjectFile
+    case invalidProjectStructure
+    case directoryNotAccessible
+    case unknownValidationFailure
+    
+    var errorDescription: String? {
+        switch self {
+        case .missingProjectFile:
+            return "This directory is not a valid Symposium project"
+        case .invalidProjectStructure:
+            return "The project structure is invalid or corrupted"
+        case .directoryNotAccessible:
+            return "Unable to access the selected directory"
+        case .unknownValidationFailure:
+            return "An unknown error occurred while validating the project"
+        }
+    }
+    
+    var recoverySuggestion: String? {
+        switch self {
+        case .missingProjectFile:
+            return "Please select a directory that contains a project.json file, or create a new project instead."
+        case .invalidProjectStructure:
+            return "Try selecting a different project directory or create a new project."
+        case .directoryNotAccessible:
+            return "Check that you have permission to access this directory and try again."
+        case .unknownValidationFailure:
+            return "Please try selecting a different directory or create a new project."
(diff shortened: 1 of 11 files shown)


nikomatsakis:
> (I had initially assumed I added symposium to an existing project)

You know...it might be nice if this worked....

.kiro/specs/project-selection-error-handling/design.md
```diff
@@ -0,0 +1,242 @@
+# Design Document
```
nikomatsakis: So--- it's a bit unclear to me how these docs are intended to fit into the workflow. Is it intentional that you commit these? And how does it fit with a project where *some* people are using Kiro and *some* are not? This document in particular seems like it is specific to this PR and could quickly become outdated.

The more "Symposium-way" would be to edit and extend the mdbook's design section.

.kiro/steering/tech.md
```diff
@@ -0,0 +1,88 @@
+# Technology Stack
+
+## Build System & Workspace
```
nikomatsakis: This seems useful enough but it feels like it should probably just be in an AGENTS.md if we don't have one already,I forget.

symposium/macos-app/Sources/Symposium/Models/ProjectManager.swift
```diff
@@ -104,7 +104,40 @@ class ProjectManager: ObservableObject, IpcMessageDelegate {
...
     deinit {
         Logger.shared.log("ProjectManager[\(instanceId)]: Cleaning up")
-        stopWindowCloseDetection()
+        // stopWindowCloseDetection() // TODO: Implement this method
+    }
+    
+
+    
+    /// Execute a process and return results
+    private func executeProcess(
```
nikomatsakis, line 113: This is good 👍

symposium/macos-app/Sources/Symposium/Models/ProjectManager.swift
```diff
@@ -501,13 +537,13 @@ class ProjectManager: ObservableObject, IpcMessageDelegate {
 
     /// Check if branch is merged into base branch
     /// 
-    /// CRITICAL: Use baseBranch directly, don't add "origin/" prefix!
-    /// getBaseBranch() already returns "origin/main", adding another "origin/" 
-    /// creates invalid "origin/origin/main" causing command failures.
+    /// Note: getBaseBranch() now returns just the branch name (e.g., "main"),
```
nikomatsakis:
```suggestion
    /// Note: getBaseBranch() returns just the branch name (e.g., "main"), 
```

I've noticed this often in LLM-generated code, it likes to write comments that assume knowledge of the code history.

symposium/macos-app/Sources/Symposium/Models/ProjectManager.swift
```diff
@@ -519,11 +555,12 @@ class ProjectManager: ObservableObject, IpcMessageDelegate {
     /// Count commits in branch that are not in base branch
     ///
     /// Uses `git rev-list --count <branch> --not <baseBranch>` to count unmerged commits.
-    /// CRITICAL: Use baseBranch directly - it already includes "origin/" prefix from getBaseBranch().
+    /// Note: getBaseBranch() now returns just the branch name, so we add "origin/" prefix.
```
nikomatsakis:
```suggestion
    /// Note: getBaseBranch() returns just the branch name, so we add "origin/" prefix.
```

symposium/macos-app/Sources/Symposium/Models/ProjectManager.swift
```diff
@@ -519,11 +555,12 @@ class ProjectManager: ObservableObject, IpcMessageDelegate {
...
     private func getUnmergedCommitCount(branchName: String, baseBranch: String, in directory: String) throws -> Int {
         let process = Process()
         process.executableURL = URL(fileURLWithPath: "/usr/bin/git")
-        process.arguments = ["rev-list", "--count", "\(branchName)", "--not", baseBranch]
+        let remoteBranch = "origin/\(baseBranch)"
```
nikomatsakis: I'm not thrilled about hardcoding origin here (and elsewhere) -- let's at least pull it into a constant somewhere.

symposium/macos-app/Sources/Symposium/Models/ProjectManager.swift
```diff
@@ -712,53 +872,89 @@ class ProjectManager: ObservableObject, IpcMessageDelegate {
     private func getBaseBranch(for project: Project) throws -> String {
         // If project specifies a default branch, use it
         if let defaultBranch = project.defaultBranch, !defaultBranch.isEmpty {
+            Logger.shared.log("ProjectManager[\(instanceId)]: Using configured default branch: \(defaultBranch)")
             return defaultBranch
         }
 
+        Logger.shared.log("ProjectManager[\(instanceId)]: No default branch configured, auto-detecting from git")
+
         // Auto-detect origin's default branch
```
nikomatsakis: Side note: we really need to have tests for this kind of logic

aaronjeline:
> The code change themselves look good. I am unclear on how `.kiro` steering files are supposed to work -- in general I've been trying to use `md` as the place to store collected documents that will guide LLMs. I'd rather not "hardcode" to Kiro, kind of goes against Symposium's unopinionated design philosophy.

Ack, shouldn't have committed these, mistake on my part, meant to put them in a gitignore

nikomatsakis (approved):
This LGTM! At the moment, from what I can see, there is no *GUI* to edit the remote name, right? But you could edit the JSON by hand I guess.

PR #33, opened 2025-10-02
Integrate GitHub MCP server functionality into Symposium

kernel-loophole:
Changes
Added as git submodule with automated binary management
Created `mcp-servers.json` for server configuration management  
Added `--git-mcp` flag to build script for one-command installation
Automatic configuration in Q CLI with GitHub token support
how to install the git MCP 
```bash
cargo setup --git-mcp
```
After that q cli add the git server and access to github tools .
supported tools in github MCP Server.
https://github.com/github/github-mcp-server?tab=readme-ov-file#available-toolsets


symposium/mcp-server/src/main.rs
```diff
@@ -73,6 +233,10 @@ enum Command {
     /// Manage persistent agent sessions
     #[command(subcommand)]
     Agent(AgentCommand),
+
+    /// Manage MCP servers
+    #[command(subcommand)]
+    Mcp(McpCommand),
```
nikomatsakis: This isn't quite what I expected -- what is the role of this subcommand?

setup/src/main.rs
```diff
@@ -220,12 +231,123 @@ fn setup_mcp_servers(binary_path: &Path) -> Result<()> {
     Ok(())
 }
 
-fn print_completion_message(built_vscode: bool, built_mcp: bool, built_app: bool) -> Result<()> {
+//Git MCP setup
+//Make sure the Q cli is installed and login
+//also need git token for repo accesses
+fn setup_git_mcp_in_q_cli() -> Result<()> {
```
nikomatsakis: OK, so here is what I am thinking.

Before we actually do the *meat* of this PR, we should begin with a refactoring. We should create a new crate, let's call it `symposium-cli-agent-util`, and it can be stored in `symposium/cli-agent-util`. We would add a `path` dependency to the `setup` crate to depend on this (and, I think, eventually from the mcp server too, but that's for some future refactoring I have in mind).

This crate would export a trait for CLI agents, something like this:

```rust
pub trait CLIAgent {
    fn name(&self) -> String;
    fn install_mcp(&self, mcp; &McpServer);
}
```

and it would export structs for all the agents we know how to manage, like this:

```rust
pub struct QCLI {
    executable_path: PathBuf,
}

impl QCLI {
    pub fn detect() -> Box<Self> {
        // ... detect `q` executable here ...
    }
}

impl CLIAgent for QCLI {
    fn name(&self) -> String {
        "Q CLI".to_string()
    }

    fn install_mcp(&self, mcp; &McpServer) {
        // ... executes `self.executable_path` with appropriate args ...
    }
}
```

And finally there should be some command to get a list of detected agents, something like this

```rust
pub fn detect_cli_agents() -> Vec<dyn CLIAgent> {
    let result: Vec<dyn CLIAgent> = vec![];

    if let Ok(agent) = QCLI::detect() {
        result.push(Box::new(agent));
    }

    if let Ok(agent) = ClaudeCode::detect() {
        result.push(Box::new(agent));
    }

    result
}
```

then we rewrite the main setup command to work with those instead of directly running executables. What do you think?

symposium/mcp-server/src/main.rs
kernel-loophole: so this as a parent command that contains nested subcommands for MCP server management.

setup/src/main.rs
kernel-loophole: That makes a lot of sense , I like the idea of starting with a refactoring step and isolating the CLI agent logic into its own crate. Having a dedicated `symposium-cli-agent-util` would definitely help clean up the setup code and make future extensions (like adding MCP integration) much smoother.
 I especially like the direction of having each agent (Q cli, ClaudeCode) implement detection and installation independently, that should make it easy to register or extend new agents later.

kernel-loophole:
> Hi @kernel-loophole! Thanks for your patience, it's been a crazy couple of days so I didn't get a chance to review until now. 😅
> 
> Anyway, I've been thinking about this, and I think the right next step is to begin with a refactoring -- take a look, what do you think.
> 
> PS, as a tip, I suggest you also ask `q` (I think you were using Q CLI, right?) to read my comments and talk it out. I have installed the `gh` command-line tool and I sometimes have to remind it to use it, although there is also a [github mcp server](https://github.com/github/github-mcp-server) that I keep meaning to try.

I've set up the GitHub MCP server, and it's working smoothly. I don’t need to manually remind it to use the server each time , it automatically invokes the  tool whenever I run a query. For example, when I ask something like “What were the most recent 5 changes?”, it correctly retrieves and responds with the latest updates.

symposium/cli-agent-util/src/lib.rs
```diff
@@ -0,0 +1,178 @@
+use anyhow::{Context, Result};
```
nikomatsakis, line 1: There seems to be two copies of this crate.

PR #35, opened 2025-10-03
Fix UI lag during taskspace deletion by making operations async

aaronjeline:
Deleting taskspaces caused main UI thread to hang. This makes a bunch of stuff in that flow async, preventing this issue.

- Made deleteTaskspace function async to prevent blocking main thread
- Added executeProcessAsync() that runs git commands in detached tasks
- Now waitUntilExit() runs off main thread completely
- Wrapped git operations (worktree removal, branch deletion) in background task
- Updated UI calls to use MainActor for thread-safe state updates
- Prevents app hanging during delete operations
- Replaced computed branchInfo property with cached state
- Load branch info asynchronously on dialog appear
- Prevents multiple git commands on every UI refresh


symposium/macos-app/Sources/Symposium/Views/ProjectView.swift
```diff
@@ -765,10 +759,16 @@ struct DeleteTaskspaceDialog: View {
...
-            deleteBranch = (branchInfo.unmergedCommits == 0 && !branchInfo.hasUncommittedChanges)
+            Task {
+                let manager = projectManager
+                let ts = taskspace
+                cachedBranchInfo = await Task.detached {
+                    manager.getTaskspaceBranchInfo(for: ts)
+                }.value
+                
+                // Set default deleteBranch toggle based on safety analysis
+                deleteBranch = (cachedBranchInfo.unmergedCommits == 0 && !cachedBranchInfo.hasUncommittedChanges)
```
nikomatsakis, line 781: I guess what happens is that, initially, the cached branch name is empty, so the user sees no checkbox at all, but then it appears some time later once the git command has completed? I think it'd be nice if we displayed a message like "...determining github branch information..." to keep them up-to-date.

symposium/macos-app/Sources/Symposium/Views/ProjectView.swift
```diff
@@ -687,7 +688,15 @@ struct DeleteTaskspaceDialog: View {
             Text("Are you sure you want to delete '\(taskspaceName)'? This will permanently remove all files and cannot be undone.")
                 .multilineTextAlignment(.center)
             
-            if !cachedBranchInfo.branchName.isEmpty {
+            if isLoadingBranchInfo {
```
nikomatsakis, line 691: ...lol, I guess you anticipated my request...

nikomatsakis (approved):
This looks good! I had noticed that UI is very slow but hadn't read carefully into it.

PR #105, opened 2026-01-15
Fix custom extensions by passing json

jackh726:
It should be *pretty* easy to remove much of the "builtin" paths here - but requires non-local ferris and cargo, so leaving that out of scope for now.

This *shouldn't* change semantic behavior (yet) if someone is still doing e.g. `--proxy cargo`.

### DIFF
--- a/src/symposium-acp-agent/src/main.rs
+++ b/src/symposium-acp-agent/src/main.rs
@@ -36,6 +36,7 @@ use clap::{Parser, Subcommand};
 use sacp::{Component, DynComponent, ProxyToConductor};
 use sacp_tokio::AcpAgent;
 use std::path::PathBuf;
+use std::str::FromStr;
 
 mod config;
 
@@ -144,16 +145,7 @@ enum RegistryCommand {
 fn build_proxies(raw_proxies: Vec<String>) -> Result<Vec<DynComponent<ProxyToConductor>>> {
     let mut proxies = Vec::with_capacity(raw_proxies.len());
     for proxy in raw_proxies {
-        let proxy = proxy.trim();
-
-        if proxy.starts_with('{') {
-            let server: sacp::schema::McpServer = serde_json::from_str(proxy)
-                .map_err(|e| sacp::util::internal_error(format!("Failed to parse JSON: {}", e)))?;
-            proxies.push(DynComponent::new(AcpAgent::new(server)));
-            continue;
-        }
-
-        anyhow::bail!("Expected json from `registry resolve-extension`.");
+        proxies.push(DynComponent::new(AcpAgent::from_str(&proxy)?));
     }
 
     Ok(proxies)


src/symposium-acp-agent/src/symposium.rs
```diff
@@ -50,49 +49,54 @@ impl SymposiumConfig {
...
-                                args: vec![],
-                            }),
-                        },
-                    };
-                    let server = crate::registry::resolve_distribution(&entry)
-                        .await
-                        .map_err(|e| sacp::Error::new(-32603, e.to_string()))?;
+        for proxy in self.proxies {
+            match proxy {
+                ProxySource::McpServer(server) => {
```
nikomatsakis: We should probably rename this. It means "an ACP proxy specified via the same format as people use for MCPs", but we will actually want to package MCP servers as proxies too, right ?

vscode-extension/src/agentRegistry.ts
```diff
@@ -338,7 +347,7 @@ export async function resolveAgentJson(agent: AgentConfig): Promise<string> {
 
   let result: McpServerJson;
 
-  if (dist.local) {
+  if (dist?.local) {
```
nikomatsakis: In theory this logic should move to the rust code right ?

src/symposium-acp-agent/src/main.rs
```diff
@@ -116,80 +129,44 @@ enum RegistryCommand {
...
+
+        if proxy.starts_with('{') {
+            let server: sacp::schema::McpServer = serde_json::from_str(proxy)
+                .map_err(|e| sacp::util::internal_error(format!("Failed to parse JSON: {}", e)))?;
+            proxies.push(DynComponent::new(AcpAgent::new(server)));
+            continue;
         }
 
-        Ok(result)
+        anyhow::bail!("Expected json from `registry resolve-extension`.");
```
nikomatsakis: fwiw, `AcpAgent::from_str` has ~this same behavior, except it also permits `foo bar`

src/symposium-acp-agent/src/registry.rs
```diff
@@ -504,12 +572,27 @@ pub fn get_binary_cache_dir(agent_id: &str, version: &str) -> Result<PathBuf> {
...
+pub async fn resolve_agent(agent: &str) -> Result<McpServer> {
+    let agent_id = if agent.starts_with('{') {
+        let entry: RegistryEntry = serde_json::from_str(agent)
+            .map_err(|e| sacp::util::internal_error(format!("Failed to parse JSON: {}", e)))?;
+
+        let resolved = resolve_distribution(&entry).await?;
+        if let Some(agent) = resolved {
+            return Ok(agent);
+        }
+        entry.id
```
nikomatsakis, line 585: This is a *little* surprising to me -- so I can do `{'id': 'foo', 'blahblah': 22}` and `foo` will be used as the name but  `'blahblah': 22` will be ignored?

src/symposium-acp-agent/src/registry.rs
```diff
@@ -521,11 +604,62 @@ pub async fn resolve_agent(agent_id: &str) -> Result<McpServer> {
...
+pub async fn resolve_extension(extension: &str) -> Result<McpServer> {
+    let ext_id = if extension.starts_with('{') {
+        let entry: RegistryEntry = serde_json::from_str(extension)
+            .map_err(|e| sacp::util::internal_error(format!("Failed to parse JSON: {}", e)))?;
+
+        let resolved = resolve_distribution(&entry).await?;
+        if let Some(agent) = resolved {
+            return Ok(agent);
+        }
+        entry.id
```
nikomatsakis, line 628: same as above; also, should we merge this logic? I guess it's simple enough

src/symposium-benchmark/src/main.rs
```diff
@@ -102,12 +104,18 @@ async fn run_benchmark(benchmark: &Benchmark, output_dir: &PathBuf) -> Result<()
...
-    let symposium = symposium_acp_agent::symposium::Symposium::new(config).with_agent(agent);
+    let config = symposium_acp_agent::symposium::SymposiumConfig::new().trace_dir(".");
+    let mut proxies = vec![];
+    for proxy in built_in_proxies()? {
+        if proxy.id == "ferris" || proxy.id == "cargo" {
+            let server = resolve_distribution(&proxy)
+                .await?
+                .ok_or_else(|| anyhow::anyhow!("Missing extension."))?;
+            proxies.push(DynComponent::new(AcpAgent::new(server)));
+        }
```
nikomatsakis, line 115: This is.. I guess... fine, it'd be nice if we didn't have to call this manually I think. (Not really saying we should edit it.)

vscode-extension/src/agentRegistry.ts
```diff
@@ -317,56 +332,17 @@ export function getCurrentAgent(): AgentConfig | undefined {
...
-      name,
-      command: "pipx",
-      args: ["run", dist.pipx.package, ...(dist.pipx.args ?? [])],
-      env: [],
-    };
-  } else {
-    throw new Error(`No compatible distribution found for agent "${agent.id}"`);
-  }
-
-  return JSON.stringify(result);
```
nikomatsakis, line 369: 🎉

src/symposium-acp-agent/src/main.rs
jackh726: Ah yes - meant to switch over to this after removing the builtin string option

src/symposium-acp-agent/src/registry.rs
jackh726: So, _yes_.

I had considered doing some "fetch and merge" behavior here if the entry provided is insufficient. But, that seemed more complicated/surprised than this.

src/symposium-acp-agent/src/registry.rs
jackh726: I'd like to merge the logic, but it's also like mildly inconvenient because of the differences between "agent" and "extension".

Probably something to just come back to.

src/symposium-benchmark/src/main.rs
jackh726: Yeah, I think the "right" thing here is to either:
1) Depend on these crates directly and use them directly (rather than going through the registry)
2) Have some `Symposium::with_default_proxies` method

src/symposium-acp-agent/src/main.rs
jackh726: Just pushed commit to switch this

PR #123, opened 2026-02-04
Add MCP server support as mods

jackh726:
Based on #122 for now

Not *entirely* sure if this is the right shape, but it works - opening just to get feedback/keep track of state

### DIFF
--- a/md/design/mods.md
+++ b/md/design/mods.md
@@ -48,8 +48,7 @@ The special value `defaults` expands to all known built-in mods:
 
 Workspace configuration can include MCP servers, which are now represented as
 mods with `kind = MCP`. They are attached directly to new sessions by the
-conductor but are stored in the workspace `mods` list. This unifies server
-configuration with the existing mod model.
+conductor.
 
 Stdio MCP servers reuse the same `ComponentSource` distribution formats as mods,
 so they can be resolved from `cargo`, `npx`, `pipx`, `binary`, `local`, or
@@ -57,7 +56,7 @@ registry entries. HTTP and SSE transports are represented using the `Http` and
 `Sse` `ComponentSource` variants and include a `name`, `url`, and optional
 headers.
 
-Local (stdio) MCP servers may now include an optional `name` field in their
+Local (stdio) MCP servers may include an optional `name` field in their
 `LocalDistribution`. When provided, this `name` is used as the MCP server's
 display name and tool prefix instead of deriving a name from the command path.
 
--- a/md/using/configuration.md
+++ b/md/using/configuration.md
@@ -32,9 +32,30 @@ Configuration
 | `1`, `2`, etc. | Toggle a mod on or off. Disabled mods stay in the list but aren't loaded. |
 | `SAVE` | Write changes to disk. They persist across sessions. |
 | `CANCEL` | Discard changes and exit. |
-| `MCPS` | Manage workspace MCP servers (add/remove stdio, http, sse servers). |
+
 
 Changes take effect immediately for the current session. Use `SAVE` to keep them for future sessions.
+
+## Local Recommendations (`RECS`)
+
+`RECS` opens an interactive menu for managing the file `config/recommendations.toml`.
+
+| Command | Description |
+|---------|-------------|
+| `ADD` | Add a new recommendation interactively. |
+| `REMOVE N` | Remove recommendation number `N` from the local recommendations list. |
+| `BACK` | Return to the main configuration menu. |
+
+The `ADD` flow currently creates a recommendation without an interactive `when` condition; by default the recommendation will apply unconditionally.
+
+## Recommendation Diffs
+
+When remote recommendations change you may be prompted with a short diff. The prompt accepts the following commands:
+
+- `SAVE` — Accept the new recommendations and apply them to the workspace.
+- `IGNORE` — Disable all newly recommended mods (keeps current config but marks the new recommendations as disabled).
+- `CONFIG` — Open the interactive configuration menu so you can selectively enable/disable or otherwise edit recommendations before saving.
+
 
 ## MCP Servers
 
--- a/src/symposium-acp-agent/src/config_agent/config_mode_actor.rs
+++ b/src/symposium-acp-agent/src/config_agent/config_mode_actor.rs
@@ -844,7 +844,7 @@ impl ConfigModeActor {
         // Commands
         msg.push_str("# Commands\n\n");
         msg.push_str("- `AGENT` - Change agent (affects all workspaces)\n");
-        msg.push_str("- `RECS` - Update local recommendations (config/recommendations.toml)\n");
+        msg.push_str("- `RECS` - Update user-defined recommendations\n");
         match mods.mods.len() {
             0 => {}
             1 => msg.push_str("- `1` - Toggle mod enabled/disabled in this workspace\n"),


src/symposium-acp-agent/src/user_config.rs
```diff
@@ -208,6 +209,50 @@ pub struct WorkspaceModsConfig {
     /// Mods with their enabled state
     #[serde(default)]
     pub mods: Vec<ModConfig>,
+    /// MCP servers configured for this workspace
+    #[serde(default)]
+    pub mcp_servers: Vec<McpServerConfig>,
```
nikomatsakis: My thinking was that we would modify `ModConfig` instead to hve a `kind` field. MCP servers would just be a *kind* of mod -- and in fact, they'd be specified I think by the same set of options (cargo/registry/whatever) as ACP mods.

md/design/mods.md
```diff
@@ -44,6 +44,26 @@ The special value `defaults` expands to all known built-in mods:
 --proxy defaults  # equivalent to: --proxy sparkle --proxy ferris --proxy cargo
 ```
 
+## MCP Servers
+
+Workspace configuration can include MCP servers, which are now represented as
+mods with `kind = MCP`. They are attached directly to new sessions by the
+conductor but are stored in the workspace `mods` list. This unifies server
+configuration with the existing mod model.
```
nikomatsakis: Nit: document the destination, not the journey -- don't say "... which are now represented ...", that references some point in the past that the reader doesn't care about. "... which are represented ...".

nikomatsakis: This applies to other parts of the docs too

md/using/configuration.md
```diff
@@ -32,9 +32,38 @@ Configuration
 | `1`, `2`, etc. | Toggle a mod on or off. Disabled mods stay in the list but aren't loaded. |
 | `SAVE` | Write changes to disk. They persist across sessions. |
 | `CANCEL` | Discard changes and exit. |
+| `MCPS` | Manage workspace MCP servers (add/remove stdio, http, sse servers). |
```
nikomatsakis: Maybe just "MCP"? (We could also accept mcps)

src/symposium-acp-agent/src/config_agent/conductor_actor.rs
```diff
@@ -249,9 +286,11 @@ async fn run_actor(
             while let Some(message) = rx.recv().await {
                 match message {
                     ConductorMessage::NewSession {
-                        request,
+                        mut request,
                         request_cx,
                     } => {
+                        request.mcp_servers.extend(mcp_servers.clone());
```
nikomatsakis, line 292: Interesting. This is not how I expected to handle this but... honestly... it's probably better than what I had in mind. I was thinking we could make a `$EXE_NAME mcp {...json...}` command that created an ACP mod which wrapped the MCP server described by `{...json...}` and then just add that into the proxy list. But that would be probably more code and also less efficient, as it would mean that MCP calls go through the proxy chain instead of being directly managed by the agent... so... yeah. This seems better. 👍

src/symposium-acp-agent/src/config_agent/config_mode_actor.rs
```diff
@@ -564,6 +567,247 @@ impl ConfigModeActor {
...
+                        "Failed to read {}: {}\n\n",
+                        local_path.display(),
+                        e
+                    ));
+                    return MenuAction::Redisplay;
+                }
+            };
+            let mut recs = match local_recs {
+                Some(recs) if !recs.mods.is_empty() => {
+                    for (m, display_index) in recs.mods.iter().zip(1..) {
```
nikomatsakis, line 595: Hmm. It seems unlikely that this code belongs *here* -- I would expect it to be a more general helper that is shared across local recommendations and recommendations from other sources, no?

src/symposium-acp-agent/src/config_agent/config_mode_actor.rs
```diff
@@ -564,6 +567,247 @@ impl ConfigModeActor {
         MenuAction::Continue
     }
 
+    /// Manage local recommendations file (`<config>/config/recommendations.toml`).
+    /// Allows listing, adding, and removing single recommendation entries.
+    async fn manage_local_recommendations(&mut self) -> MenuAction {
```
nikomatsakis, line 572: I've got a side branch doing something similar, but different. My branch is adding search capabilities to search crates.io and so forth and let people pick from the results.

src/symposium-acp-agent/src/config_agent/config_mode_actor.rs
```diff
@@ -582,10 +826,16 @@ impl ConfigModeActor {
         } else {
             for (m, display_index) in mods.mods.iter().zip(1..) {
                 let name = m.source.display_name();
+                let mcp = matches!(m.kind, ModKind::MCP)
+                    .then_some(" (MCP)")
+                    .unwrap_or("");
```
nikomatsakis, line 831: Yep, as expected, this ought to be pulled into some kind of helper function.

Also, unrelated, but we should be including the "why recommended" string here...

src/symposium-acp-agent/src/config_agent/config_mode_actor.rs
```diff
@@ -594,6 +844,7 @@ impl ConfigModeActor {
         // Commands
         msg.push_str("# Commands\n\n");
         msg.push_str("- `AGENT` - Change agent (affects all workspaces)\n");
+        msg.push_str("- `RECS` - Update local recommendations (config/recommendations.toml)\n");
```
nikomatsakis: I feel like users don't understand what this means. It feels like these are "per-user mods", right?

src/symposium-acp-agent/src/main.rs
```diff
@@ -306,22 +306,22 @@ async fn main() -> Result<()> {
...
 
-            let mods = if no_mods {
+            let recs = if no_mods {
                 vec![]
             } else {
                 // Load recommendations from remote (with cache fallback)
                 let recommendations =
                     remote_recommendations::load_recommendations(&config_paths).await?;
-                recommendations.for_workspace(&workspace).mod_sources()
+                recommendations.for_workspace(&workspace).mods
```
nikomatsakis, line 315: I don't really follow what's going on here-- just some renaming I guess and "cleanup" I guess?

src/symposium-acp-agent/src/recommendations.rs
```diff
@@ -265,39 +264,28 @@ pub struct WorkspaceRecommendations {
 }
 
 impl WorkspaceRecommendations {
-    /// Get all mod sources as a set (for diffing with config)
-    pub fn mod_sources(&self) -> Vec<ComponentSource> {
-        self.mods.iter().map(|r| r.source.clone()).collect()
-    }
```
nikomatsakis, line 271: ...looks like it. ok

md/design/mods.md
jackh726: Tbh this is all some LLM text. I had that it does this. (Like put comments places like "we no longer do this")

src/symposium-acp-agent/src/config_agent/config_mode_actor.rs
jackh726: Hmm, like, the display code? Or what exactly are you expecting to be shared? (and used for what?)

src/symposium-acp-agent/src/config_agent/config_mode_actor.rs
jackh726: I don't think it's incompatible :) The current `ADD` flow is pretty simple, but could be extended with search functionality.

src/symposium-acp-agent/src/config_agent/config_mode_actor.rs
jackh726: Hmm, I guess I could see some "mod explanation" that is name + type. I think for now the duplication is okay, because it's kind of "specialized", but I can get to this with a "why recommended" string later?

src/symposium-acp-agent/src/config_agent/config_mode_actor.rs
jackh726: I'm updating this to "Update user-defined recommendations" for now

md/design/mods.md
nikomatsakis: @jackh726 I assumed it was LLM text -- I've found by default that it tends to write things that assume the person reading the comment is aware of the prompt conversation that led to the changes. If you think it's appropriate in this case, that's fine, but I feel like docs should usually just "document the destination, not the journey", as I said.

src/symposium-acp-agent/src/config_agent/config_mode_actor.rs
nikomatsakis: I meant the logic to display "(MCP)" and so forht. IT's fine like this, I think there is cleanup we could do later.

PR #168, opened 2026-04-10
feat: remind agent to run cargo fmt after Rust files change

gme-muriuki:
Add `cargo_fmt.rs` with mtime-based detection of `.rs` file changes. After each `PostToolUse`, Symposium compares current file modification times against a snapshot stored in session state and injects a suggestion to run `cargo fmt` if any files changed. The reminder frequency is configurable via `fmt-reminder` under `[hooks]` in `config.toml` (once, always, never; default: once).

Also fixes hook dispatch to use a platform-appropriate shell command (`cmd /C` on Windows, `sh -c` elsewhere) via a new `shell_command()` helper.

Changes:
- Add `cargo_fmt.rs` with change detection and suggestion logic
- Extend `SessionData` with `rust_file_snapshot` and `rust_fmt_reminder_sent`
- Add `FormatReminderPolicy` enum and `remind_format_policy` to `HooksConfig`
- Wire `maybe_suggest_rust_fmt` into `handle_post_tool_use` in `hook.rs`
- Add `shell_command()` helper to `hook.rs` for cross-platform dispatch
- Add `cargo-fmt-reminder-flow.md` documenting the new flow
- Update `important-flows.md`, `session-state.md`, and `module-structure.md`

## What does this PR do?

Implements the cargo fmt reminder discussed in the sync meeting.

<details>

**AI disclosure.**

* The AI tool authored small parts of the code

**Confidence level.**

* It seems reasonable to me, I'd like to know what others think

**Questions for reviewers.**

- Should `Once` eventually be configurable as "every N tool uses" rather than strictly once per session?

</details>


src/cargo_fmt.rs
```diff
@@ -0,0 +1,320 @@
...
+/// and compared at the start of the next one.
+pub fn snapshot_rust_files(cwd: &Path) -> HashMap<PathBuf, SystemTime> {
+    let mut mtimes = HashMap::new();
+    collect_rust_file_mtimes(cwd, &mut mtimes);
+    mtimes
+}
+
+/// Walk `dir` recursively, collecting mtimes(modification times) of all `*.rs` files.
+fn collect_rust_file_mtimes(dir: &Path, mtimes: &mut HashMap<PathBuf, SystemTime>) {
+    let entries = match std::fs::read_dir(dir) {
```
nikomatsakis: I would usually use https://crates.io/crates/walkdir for this; I suspect there are surprising edge cases (e.g., I don't know, symbolic links that form a cycle or something nutso like that)

src/cargo_fmt.rs
```diff
@@ -0,0 +1,320 @@
...
+/// under `cwd`. Stored in session state at the end of each `PostToolUse`
+/// and compared at the start of the next one.
+pub fn snapshot_rust_files(cwd: &Path) -> HashMap<PathBuf, SystemTime> {
+    let mut mtimes = HashMap::new();
+    collect_rust_file_mtimes(cwd, &mut mtimes);
+    mtimes
+}
+
+/// Walk `dir` recursively, collecting mtimes(modification times) of all `*.rs` files.
+fn collect_rust_file_mtimes(dir: &Path, mtimes: &mut HashMap<PathBuf, SystemTime>) {
```
nikomatsakis: I would use a `BTreeMap` for deterministic ordering, or else some deterministic hash function like https://crates.io/crates/fxhash

src/cargo_fmt.rs
```diff
@@ -0,0 +1,320 @@
...
+    // Any new or modified files?
+    for (path, mtime) in &current {
+        match previous.get(path) {
+            Some(prev) if prev != mtime => return true, // modified
+            None => return true,                        // new file
+            _ => {}
+        }
+    }
+
+    // Any deleted files?
```
nikomatsakis: Let's just check up-front:

```
if previous.len() != current.len() { }
```

src/hook.rs
```diff
@@ -1093,4 +1098,50 @@ mod tests {
...
+
+        let output = handle_post_tool_use(&sym, &payload).await;
+
+        // 5. Verify the reminder is sent in the HookOutput
+        let specific = output.hook_specific_output.unwrap();
+        assert_eq!(specific.hook_event_name, "PostToolUse");
+        let context = specific.additional_context.unwrap();
+        assert!(context.contains("cargo fmt"));
+        assert!(context.contains("modified"));
+    }
```
nikomatsakis: can we add an integration test?

https://symposium.dev/design/test-harness.html

this would let us test the configuration settings. I imagine we would invoke the hooks manually -- i.e., check that we get it once, then change files, and check we don't get it again, that sort of thing.

nikomatsakis (changes requested):

src/cargo_fmt.rs
Gme-muriuki: Thanks for the suggestion! I wasn’t aware of walkdir before. It sure does seem like the right abstraction for this kind of task. I’ll switch to using walkdir for this and update the PR

src/hook.rs
Gme-muriuki: Yeah, of course. 👍 I'll add them too

src/cargo_fmt.rs
Gme-muriuki: I did opt for `BTreeMap` since it's like `fxhash` crate is unmaintained.

src/cargo_fmt.rs
```diff
@@ -0,0 +1,112 @@
...
+) -> Option<String> {
+    // Optimization: If the policy is never, don't bother walking the file system.
+    if matches!(policy, FormatReminderPolicy::Never) {
+        return None;
+    }
+
+    let changed = rust_files_changed_since(cwd, &session.rust_file_snapshot);
+
+    // Refresh snapshot for the next tool use
+    session.rust_file_snapshot = snapshot_rust_files(cwd);
```
nikomatsakis:
```suggestion
    // Check if any Rust files have changed.
    let current = snapshot_rust_files(cwd);
    if session.rust_file_snapshot == current {
        return None;
    }
    session.rust_file_snapshot = current;
```

src/cargo_fmt.rs
```diff
@@ -0,0 +1,112 @@
...
+    for (path, mtime) in &current {
+        match previous.get(path) {
+            Some(prev) if prev != mtime => return true, // modified
+            None => return true,                        // new file
+            _ => {}
+        }
+    }
+
+    false
+}
```
nikomatsakis:
```suggestion
```

I don't think we need this function at all (see below)

src/cargo_fmt.rs
nikomatsakis: Actually we could do this a bit better, right -- probably the first thing we should do is match on the policy. If the policy is "once" and we've already sent the reminder, no need to look at the disk again.

tests/cargo_format.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    let subdir = root.join("sub");
+    fs::create_dir(&subdir).unwrap();
+
+    let rs1 = make_temp_rs(root, "a.rs");
+    let rs2 = make_temp_rs(&subdir, "b.rs");
+    let txt = root.join("ignore.txt");
+    fs::write(&txt, b"not rust").unwrap();
+
+    let snapshot = snapshot_rust_files(root);
```
nikomatsakis: Hmm, I don't think this should be part of the public interface we export. 

Honestly I wouldn't bother with unit testing this function, it seems straightforward enough.

tests/cargo_format.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    assert!(snapshot.contains_key(&rs1));
+    assert!(snapshot.contains_key(&rs2));
+    assert!(!snapshot.contains_key(&txt));
+}
+
+#[test]
+fn collect_rust_file_mtimes_ignores_unreadable_directories() {
```
nikomatsakis: yeah I don't think this test adds much either

src/lib.rs
```diff
@@ -1,4 +1,5 @@
 pub mod agents;
+pub mod cargo_fmt;
```
nikomatsakis: I don't think this should be public

src/lib.rs
```diff
@@ -9,11 +10,11 @@ pub mod init;
 pub mod mcp;
 pub mod output;
 pub mod plugins;
+pub mod session_state;
```
nikomatsakis: nor this

tests/cargo_format.rs
```diff
@@ -0,0 +1,242 @@
...
+    fs::write(&rs_path, b"// modified code").unwrap();
+
+    // 4. Simulate PostToolUse event (e.g., after a Bash command);
+    let payload = PostToolUsePayload {
+        tool_name: "Bash".to_string(),
+        tool_input: serde_json::json!({"command": "echo done"}),
+        tool_response: serde_json::json!({"exit_code": 0}),
+        session_id: Some(session_id.to_string()),
+        cwd: Some(root.to_str().unwrap().to_string()),
+    };
```
nikomatsakis: So *this* is closer to what I expect from an integration test. I'd not bother with the other tests. I want to test the hook flow. Basically what I expect is a set of tests that test out the various configuration and which have this structure

- create files
- invoke post-tool-use hook 
- modify file
- invoke post-tool-use hook 
- maybe observe advice to run fmt (depends on cifug)
- modify file
- invoke post-tool-use hook 
- maybe observe advice to run fmt (depends on cifug)

we first want to test

- fmt disabled (never observe)
- fmt set to 'once" (observe once)
- fmt set to 'awlays (observe twice)

then we want tests for different kinds of modifications (modifying file, creating file, deleting file).

nikomatsakis: The integration test harness in https://symposium.dev/design/test-harness.html -- I don't think you're fully using it. I expect these tests to look like

```rust
let ctx = with_fixture(&["plugins0"]);

// make files

ctx.hook(PostToolUseHook { .. })

// modify files 
...
```

something like that.

nikomatsakis (changes requested):

src/cargo_fmt.rs
Gme-muriuki: Sure we don't. I've fixed it 👍

src/cargo_fmt.rs
Gme-muriuki: Totally 💯.

tests/cargo_format.rs
Gme-muriuki: Good point. Integration test for the hook flow will cover that sufficiently

tests/cargo_format.rs
Gme-muriuki: Hmm, I see. Working on it. 👍 halfway done.
And the integration test are slow after adding the `with_fixture(&["my_fixture" , "workspace0"])`, is that expected. Without "workspace0" they were pretty fast.

Gme-muriuki: Oh, my bad. I was coding them and testing them offline, that's why they were slow. Apparently this `symposium-recommendations = true` needs me to be online 👍

Gme-muriuki:
It appears like the changes to be introduced by PR #170 are introducing breaking changes for the `invoke_hook` function. I've updated  the integration tests but when PR #170  lands I will update them to align with the new test harness .

I'm not sure whether we should merge this if ready, or we should wait for the other pr to land and have it fixed first then merged...

 In the meantime, the current tests cover the core behaviors (I could think off) and I'm happy to address any other review comments.

src/cargo_fmt.rs
```diff
@@ -0,0 +1,85 @@
...
+    policy: &FormatReminderPolicy,
+) -> Option<String> {
+    if matches!(policy, FormatReminderPolicy::Once) && session.rust_fmt_reminder_sent {
+        return None;
+    }
+    if matches!(policy, FormatReminderPolicy::Never) {
+        return None;
+    }
+
+    let current = snapshot_rust_files(cwd);
```
jackh726, line 59: I'm wondering if we should do a followup check after this: `cargo fmt -- --check`?

This would mean that the false-positive rate is less, but does mean we are essentially duplicating the work. I think that's okay.

This will also mean we don't need to set this option to `never` in the tests.

src/hook.rs
```diff
@@ -728,7 +746,8 @@ mod tests {
     #[tokio::test]
     async fn builtin_post_tool_use_returns_empty_for_now() {
         let tmp = tempfile::tempdir().unwrap();
-        let sym = Symposium::from_dir(tmp.path());
+        let mut sym = Symposium::from_dir(tmp.path());
+        sym.config.hooks.remind_format_policy = crate::config::FormatReminderPolicy::Never;
```
jackh726, line 750: Hmm, I don't like setting this in the test. I have a couple solutions:
1) Split out a `make_test_symposium` function that takes a Path and returns a `Symposium`. This at least separates out the logic.
2) *Always* run `cargo fmt -- --check` before suggesting.
3) Run `cargo fmt -- --check` only on the *first* change, just to prevent false positives.

src/cargo_fmt.rs
Gme-muriuki: Hmm, I like the idea. It does eliminate false positives entirely. I'm  happy to implement it as a follow up PR to keep this one focused. What do you think?

src/hook.rs
Gme-muriuki: Solution 1 looks pretty straight forward. I can fix that up, then work on the `cargo fmt -- --check` as a separate PR.

PR #172, opened 2026-04-14
Add installation and hook source infra

jackh726:
## What does this PR do?

This adds the necessary infra to actually download/install the things for hooks.

<details>

**AI disclosure.**

**Confidence level.**

* Early draft, want feedback
* It seems reasonable to me, I'd like to know what others think

</details>


src/hook.rs
```diff
@@ -527,15 +527,47 @@ pub(crate) fn dispatch_plugin_hooks<E: AgentHookEvent>(
...
+                let resolved = runtime.block_on(resolve_distribution(sym, distribution));
+                let resolved_path = match resolved {
+                    Ok(p) => p,
+                    Err(e) => {
+                        tracing::error!(error = %e, "failed to resolve hook distribution");
+                        continue;
+                    }
+                };
+
+                Command::new("sh")
```
nikomatsakis, line 559: we have this more than once, we should extract a helper for this

src/hook.rs
```diff
@@ -527,15 +527,47 @@ pub(crate) fn dispatch_plugin_hooks<E: AgentHookEvent>(
...
+                    .arg(&resolved_path.path)
+                    .args(&resolved_path.args)
+                    .stdin(Stdio::piped())
+                    .stdout(Stdio::piped())
+                    .stderr(Stdio::piped())
+                    .spawn()
+            }
+            (None, Some(command)) => Command::new("sh")
```
nikomatsakis, line 567: here, but also elsewhere in the codebase, like the hooks code

src/hook.rs
```diff
@@ -648,6 +650,9 @@ fn hooks_for_payload(
         let name = plugin.name.clone();
         for hook in &plugin.hooks {
             tracing::debug!(?hook);
+            if hook.agent.map_or(false, |only_agent| only_agent != agent) {
```
nikomatsakis, line 660: hmm, i'm not sure what this is trying to do

resources/plugins/rtk.toml
```diff
@@ -10,3 +10,9 @@ source = "git"
 repo = "https://github.com/rtk-ai/rtk.git"
 commit = "master"
 path = "hooks/claude/rtk-rewrite.sh"
+
+[[hooks.requirements]]
+type = "cargo"
+crate_name = "rtk"
+version = "latest"
+git = "https://github.com/rtk-ai/rtk"
```
nikomatsakis, line 18: what does it mean to specify both here? I wonder if this should be 

```toml
[[hooks.requirements]]
cargo = { crate_name = "rtk", version = "latest" }
```

nikomatsakis: Ok, I see, we'll use git. Got it.

nikomatsakis: I am actually surprised that this works, I thought we would need a `cargo = ` since this is coming from an enum

src/installation.rs
```diff
@@ -0,0 +1,59 @@
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+
+use crate::config::Symposium;
+use crate::distribution::{get_binary_cache_dir, install_cargo_crate, query_crate_binaries};
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(tag = "type", rename_all = "lowercase")]
```
nikomatsakis, line 8: oh, maybe it's the `tag = "type"`?

src/init.rs
```diff
@@ -90,6 +94,10 @@ fn resolve_project_agents(opts: &InitOpts, existing: &[AgentEntry]) -> Result<Ve
 pub async fn init_user(sym: &mut Symposium, out: &Output, opts: &InitOpts) -> Result<()> {
     out.println("Setting up symposium for your user account.\n");
 
+    if !opts.no_default_plugins {
+        tokio::fs::write(sym.config_dir().join("plugins").join("rtk.toml"), RTK_PLUGIN).await?;
```
nikomatsakis, line 98: I expected this to be in https://github.com/symposium-dev/recommendations

src/init.rs
```diff
@@ -10,13 +10,17 @@ use crate::agents::Agent;
 use crate::config::{AgentEntry, ProjectConfig, Symposium};
 use crate::output::{Output, display_path};
 
+const RTK_PLUGIN: &str = include_str!("../resources/plugins/rtk.toml");
```
nikomatsakis, line 13: was this file added?

src/hook.rs
jackh726: Hmm, not sure where else you mean?

I'm also not entirely sure what extracting a helper would look like here: it's not even like the args are the same here.

md/crate-authors/creating-a-plugin.md
```diff
@@ -36,8 +36,17 @@ source.path = "skills/v1"
...
+# simple local script
 command = "./scripts/validate.sh"
 
+[[hooks]]
+name = "cargo-installer-hook"
+event = "PreToolUse"
+# use a cargo-distributed binary (example)
+distribution = { source = "cargo", crate = "ripgrep", version = "13.0.0", binary = "rg" }
+# install a helper tool before running the hook
+requirements = [ { type = "cargo", crate = "my-tool", version = "0.2.1" } ]
```
nikomatsakis: I feel confused about what these mean :)

md/reference/plugin-definition.md
```diff
@@ -63,6 +67,89 @@ Each `[[skills]]` entry declares a group of skills.
...
+
+```toml
+[[installations]]
+name = "agent-hook"
+source = "binary"
+"linux-x86_64"   = { archive = "https://example.com/linux.tar.gz",  cmd = "./hook" }
+"darwin-aarch64" = { archive = "https://example.com/macos.tar.gz", cmd = "./hook" }
+```
+
+Symposium selects the entry matching the current platform (`darwin-aarch64`, `darwin-x86_64`, `linux-x86_64`, `linux-aarch64`, `windows-x86_64`, etc.), downloads and extracts the archive once, and runs `cmd`.
```
nikomatsakis: we should document where the archive is extracted to, similar to the `cargo` entry?

md/reference/plugin-definition.md
```diff
@@ -71,15 +158,126 @@ Each `[[hooks]]` entry declares a hook that responds to agent events.
...
+[[hooks]]
+name = "check"
+event = "PreToolUse"
+command = { source = "local", command = "scripts/check.sh" }
+args = ["--strict"]
+```
+
+### Requirements
+
+`requirements` ensures other installations are acquired before the hook runs. Useful when the hook's command relies on something else being on disk (or eventually on `$PATH`).
```
nikomatsakis, line 248: A thought: if you have a requirement, we should set two environment variables for the hook's command, something like `$INSTALLATION_DIR` (identifies the directory where file was downloaded) and `$INSTALLATION` identifies the binary.

This way you can reference the downloaded binary.

nikomatsakis: I guess our plan was to put those binary directories on the PATH. This text makes it sound like this is not part of this PR?

tests/fixtures/plugin-hooks0/dot-symposium/plugins/test-plugin/SYMPOSIUM.toml
```diff
@@ -0,0 +1,43 @@
...
+matcher = "Read"
+command = "named-shell"
+format = "symposium"
+
+# Hook with a (no-op) requirement — verifies the requirements path runs.
+[[hooks]]
+name = "with-requirements"
+event = "PreToolUse"
+matcher = "Edit"
+requirements = [{ source = "shell", command = "true" }]
```
nikomatsakis: how is it "observable" that the requirements path runs?

nikomatsakis (approved):
Beautiful. Approved on my end, left a question or two.

md/reference/plugin-definition.md
jackh726: Yes, anything around modifying the PATH, etc. isn't in this PR - I will do it in a followup. I'll add the INSTALLATION_DIR and INSTALLATION vars with that.

tests/fixtures/plugin-hooks0/dot-symposium/plugins/test-plugin/SYMPOSIUM.toml
jackh726: Actually, it doesn't. This requirement is actually kind of bogus (but allowed, since "shell" installations are just no-ops in terms of actual installation steps).

But, it does bring up a good point: there needs to be a (clear) distinction between "run this command during installation" vs "run this command as the executable"; the current code does the latter always.

I'm gonna adjust this, and also error on the latter for inline requirements.

jackh726:
I made some more changes:
- Added an `installation_steps` field to run after cargo install or github clone
- Split `path` into `executable` and `script`
- Commands can be added without a source, and *just* an executable or a script
- Remove `local`/`shell` in favor of the above
- Remove the `binary` installation for now

PR #188, opened 2026-04-21
fix(hooks): match tool hooks with regex

anaslimem:
## What does this PR do?
Fixes hook matcher evaluation for Claude Code and Gemini tool hooks by treating `matcher` as a regex applied to the tool name instead of a substring check. This restores support for patterns like `mcp__.*` and `^Bash$`, and adds regression tests for regex matching and invalid patterns.
<details>

**AI disclosure.**

* I used an AI tool for research, autocomplete

**Confidence level.**

* I am very happy with it


nikomatsakis (approved):
This PR looks good. Reading the Claude Code docs though I cannot for the life of me tell what the intended semantics are. This seems @jackh726 to relate to the idea that if we install "claude hooks" directly into claude's configuration, we don't have to emulate it.

jackh726 (approved):

anaslimem:
@nikomatsakis  I think  Claude-native hooks can rely on Claude’s matcher semantics. Symposium still needs correct matcher handling for the hooks it dispatches internally. What do u think

PR #202, opened 2026-05-08
fix: reject invalid SKILL.md YAML frontmatter

jlizen:
## Problem

Symposium parsed `SKILL.md` frontmatter with a loose line-based parser, so malformed YAML like `description: [Critical] ...` could pass validation and be installed by hook-managed sync. Agents then rejected the copied skill at startup because they parse the same frontmatter as YAML.

We are also serving some broken frontmatter in `symposium/recommendations`. That package's CI will now fail due to our tighter validation, and I'll fix it over there (https://github.com/symposium-dev/recommendations/pull/9)

## Solution

Parse skill frontmatter as YAML before matching or installing skills, require metadata fields to be strings, and add a sync regression test proving invalid skill frontmatter is skipped instead of copied into `.agents/skills`.

symposium will warn when it skips loading due to misformatted files. That might not be the right ux choice long-term, but whilst most of our users are people also developing symposium, it's probably good :)

I didn't add a snapshot test since the sync output is not designed to have i/o injected and it would have exploded the diff. Happy to follow up with that if desired.

Here's a manual test though:
```
 Running `/Users/jess/workplace/symposium/target/debug/cargo-agents sync`
⚠️  skipping /tmp/symposium-home/plugins/bad-skill/SKILL.md: failed to load standalone skill: failed to parse frontmatter in /tmp/symposium-home/plugins/bad-skill/SKILL.md
ℹ️  scanning 2 workspace dependencies
🟢 /private/tmp/symposium-workspace/.codex/hooks.json: hooks already registered
➖ removed skill rust-best-practice from /private/tmp/symposium-workspace/.agents/skills/rust-best-practice
ℹ️  no applicable skills found for workspace dependencies
```

**AI disclosure.**

**Confidence level.**

* I am very happy with it

### DIFF
--- a/src/skills.rs
+++ b/src/skills.rs
@@ -768,6 +768,7 @@ mod tests {
                 plugin,
             }],
             standalone_skills: vec![],
+            warnings: vec![],
         };
 
         // Query for serde - should find no skills because plugin doesn't apply
@@ -807,6 +808,7 @@ mod tests {
                 plugin,
             }],
             standalone_skills: vec![],
+            warnings: vec![],
         };
 
         // Query for serde - should find no skills because group doesn't match
@@ -867,6 +869,7 @@ mod tests {
                 plugin,
             }],
             standalone_skills: vec![],
+            warnings: vec![],
         };
 
         // Query for serde - should find the skill because all levels match
@@ -955,6 +958,7 @@ mod tests {
         let registry = PluginRegistry {
             plugins: Vec::new(),
             standalone_skills: vec![skill],
+            warnings: vec![],
         };
 
         let sym = crate::config::Symposium::from_dir(tmp.path());
--- a/src/skills.rs
+++ b/src/skills.rs
@@ -354,6 +354,7 @@ fn collect_skills_applicable_to(
 
 /// Raw frontmatter fields extracted from a SKILL.md file.
 /// `crates` is comma-separated on a single line.
+#[derive(Debug)]
 struct RawFrontmatter {
     fields: BTreeMap<String, String>,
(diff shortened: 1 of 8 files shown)


src/skills.rs
```diff
@@ -385,24 +386,33 @@ fn parse_frontmatter(content: &str) -> Result<RawFrontmatter> {
...
 
-    for line in frontmatter_text.lines() {
-        let line = line.trim();
-        if line.is_empty() {
+    for (key, value) in mapping {
+        let Some(key) = key.as_str() else {
+            bail!("SKILL.md frontmatter keys must be strings");
+        };
+
+        if key == "applies-when" {
```
nikomatsakis: we should just remove this `if`, I thikn it .. serves no purpose?

tests/init_sync.rs
```diff
@@ -120,6 +120,36 @@ async fn sync_installs_skills() {
...
 
+/// `sync` rejects malformed skill frontmatter before installing skills.
+#[tokio::test]
+async fn sync_skips_invalid_skill_frontmatter() {
+    with_fixture(
+        TestMode::SimulationOnly,
+        &["invalid-skill0", "workspace0"],
+        async |mut ctx| {
+            ctx.symposium(&["init", "--add-agent", "codex"]).await?;
+            ctx.symposium(&["sync"]).await?;
```
nikomatsakis: I'd like a test that `plugin validate` will reject a plugin repo with a standalone skill that is illformed

nikomatsakis (changes requested):
Seems good

nikomatsakis (approved):

PR #203, opened 2026-05-08
fix(clippy): nested ifs/let chains

jlizen:
## What does this PR do?

Initial pass to address clippy lints, before adding to to CI.

This one has a few because they would have caused merge conflicts otherwise:
- Collapse nested ifs into let chains 
- manual_flatten
- dead_code
- option_as_ref_deref

All seem reasonable to me

<details>

**AI disclosure.**

**Confidence level.**

* I am very happy with it

</details>

### DIFF
--- a/src/installation/git.rs
+++ b/src/installation/git.rs
@@ -219,62 +219,56 @@ impl GitCacheManager {
         let meta_path = plugin_dir.join(CACHE_META_FILENAME);
 
         // If cached, check freshness according to update level
-        if plugin_dir.exists() {
-            if let Some(meta) = self.read_meta(&meta_path) {
-                // Debounce: if fetched recently, skip the API call entirely
-                if matches!(update, UpdateLevel::None) {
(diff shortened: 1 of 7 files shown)


src/installation/git.rs
```diff
@@ -219,62 +219,56 @@ impl GitCacheManager {
...
                         &meta_path,
-                        &sha,
+                        &remote_sha,
                     )
                     .await?;
                     return Ok(plugin_dir);
                 }
-
-                // None (past debounce) or Check: check freshness via API
-                match self.client.resolve_commit_sha(source).await {
```
jlizen, line 251: @jackh726 this does indeed look like dead code, but is it supposed to be?

nikomatsakis (approved):
All these changes look good to me. Exciting that `if foo && let` is stable now, I keep forgetting. But before I merge, I'd like to hear from @jackh726 on that dead code question.

src/installation/git.rs
jackh726: *shrug* idk, I think Niko (or Claude) wrote this code.

PR #223, opened 2026-05-14
Add global install option for cargo and add env vars for binaries and directories

jackh726:
## What does this PR do?

Two main things here:
- Add a `global` option to installs of `source = "cargo"`, where we don't use `--root` and instead install into the user's PATH
- Add `SYMPOSIUM_<name>` and `SYMPOSIUM_DIR_<name>` environment variables for the executable (or script) and the install directory for scripts to be able to use

<details>

**AI disclosure.**

</details>

### DIFF
--- a/md/reference/plugin-definition.md
+++ b/md/reference/plugin-definition.md
@@ -191,7 +191,7 @@ git = "https://github.com/example/tool"
 executable = "tool"   # required for git sources (crates.io is not consulted)
 ```
 
-To install into the user's global cargo location (`~/.cargo/bin`) instead of a symposium-managed cache, set `global = true`. No `--root` is passed; `$PATH` is not augmented (the binary is expected to already be on `$PATH`).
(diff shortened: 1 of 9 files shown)


nikomatsakis (approved):
This looks cool! I like how this installation infra is shaping up.

md/reference/plugin-definition.md
```diff
@@ -132,6 +132,17 @@ git = "https://github.com/example/tool"
 executable = "tool"   # required for git sources (crates.io is not consulted)
 ```
 
+To install into the user's global cargo location (`~/.cargo/bin`) instead of a symposium-managed cache, set `global = true`. No `--root` is passed; `$PATH` is not augmented (the binary is expected to already be on `$PATH`).
```
nikomatsakis: Should we explain *when* you might want that? I guess it is intended for cases where the user might want to use the CLI too?

nikomatsakis: Do we expect this to change auto-update behavior? (I think no)

nikomatsakis (commented):
This looks cool! I like how this installation infra is shaping up.

PR #247, opened 2026-06-24
fix(copilot): Use `settings.json` for hooks config

### LINKED ISSUE #222: Global hooks for copilot should probably be added to `settings.json`, not `config.json`
See https://docs.github.com/en/copilot/reference/copilot-cli-reference/cli-config-dir-reference#settingsjson

`config.json` is now automatically managed and shouldn't be edited manually. User-editable settings should go in `settings.json`.

If global hooks are added to `config.json`, they are automatically moved to `settings.json` the next time the GH copilot cli runs. Moreover, some comments are added at the top of the file (a la JSONC) which trip up `serde_json`.

abusch:
## What does this PR do?

Hooks settings for Github copilot are now stored in the user settings file (`~/.copilot/settings.json`) and not in the global config file (`~/.copilot/config.json`, which is not meant to be directly edited).

close #222

<details>

**AI disclosure.**

* Other: No AI was used for this PR.

</details>

### DIFF
--- a/src/agents/mod.rs
+++ b/src/agents/mod.rs
@@ -166,7 +166,7 @@ impl Agent {
             }
             Agent::Codex => register_codex_hooks(&home.join(".codex").join("hooks.json"), out),
             Agent::Copilot => {
-                register_copilot_hooks_global(&home.join(".copilot").join("config.json"), out)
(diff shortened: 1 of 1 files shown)


nikomatsakis (approved):
Thanks!

PR #251, opened 2026-06-29
Add basic telemetry infra

jackh726:
(based on #241)

## What does this PR do?

- Opt-in telemetry infrastructure: a new local, anonymous JSONL event log; with config, daily roll-off/30-day retention, and a cargo agents telemetry subcommand (status/show).
- init telemetry opt-in prompt and supporting config plumbing in config.rs/cli.rs.
- New Stop hook event

<details>
<summary>Disclosure questions</summary>

**AI disclosure.**

* The AI tool authored large parts of the code

</details>

PR #261, opened 2026-07-10
Telemetry/243 emitting events

### LINKED ISSUE #243: Telemetry: Emitting events
Wire up the places that should produce telemetry events.

Part of the telemetry initiative — see sync discussions: [June 9](https://symposium-dev.zulipchat.com/#narrow/channel/530148-general/topic/Symposium.20sync/with/601508415), [June 23](https://symposium-dev.zulipchat.com/#narrow/channel/530148-general/topic/2026.2E06.2E23.20-.20Sync/with/606060514).

## Milestones

- [ ] **Add "Stop" / end-of-turn hook event** — New hook event fired when the agent signals it has completed a turn. Initially minimal payload ("turn ended"); later attach summaries (files edited, tokens used, etc.).
- [ ] **Activity metric events** — Emit events for: skill activation (which skill, which crate triggered it), hook invocation (which hook, which plugin, duration), sync runs (skills installed/reaped, plugins matched), session start (agent type, workspace crate count).
- [ ] **Outcome-oriented events (experimental)** — "Churn" signal: detect repeated edits to the same file/region within a turn. Crate-correlated churn: tag churn events with which dependency is involved. Best-effort/heuristic, will need iteration.

## Notes

- The Stop event is essential for the feedback collection workstream too.
- Activity metrics are straightforward; outcome metrics require experimentation.

gme-muriuki:
## What does this PR do?

closes #243 

Makes the opt-in telemetry log (added in #251) record real activity. The hook pipeline now emits 8 anonymous event kinds (counts and coarse metadata only, no prompt text or file paths): 

- the wire funnel (session start, prompt, tool use, stop), the sync pass (which plugins and skills activated, and the crates that triggered them), and hook dispatch. 
- Producers return plain data, a pure mapping turns it into events, and `Symposium::record_telemetry` is the single place `[telemetry] enabled` is checked. Off by default, local JSONL, nothing uploaded.

<details>

**AI disclosure.**

**Questions for reviewers.**

</details>


src/hook.rs
```diff
@@ -273,6 +279,18 @@ pub async fn execute_hook(
...
 
+        // Emit telemetry for this pass. A blocking hook (exit 2 / killed) returned
+        // above via `?`, so a blocked pass is never reached here and records no
+        // telemetry: the block path stays free of best-effort work.
+        let events = telemetry_events(agent, &sym_input, sync_summary, hook_invocations);
+
+        if !events.is_empty() {
+            let session_id = sym_input.session_id().map(str::to_string);
+            for ekind in events {
+                sym.record_telemetry(session_id.clone(), ekind);
```
nikomatsakis: would it be more efficient (prob doesn't matter...but even so...) to pass in the whole array?

src/hook.rs
```diff
@@ -282,6 +300,93 @@ pub async fn execute_hook(
...
+                    plugin: pa.name,
+                    crates: pa.crates,
+                }),
+        );
+
+        events.extend(
+            summary
+                .skills
+                .into_iter()
+                .map(|ska| EventKind::SkillActivation {
```
nikomatsakis, line 386: I would not call this an *activation* but rather an *installation* or *discovery*. 

To me, a skill *activation* is when the model actually reads the skill, versus simply having it available.

src/hook.rs
```diff
@@ -282,6 +300,93 @@ pub async fn execute_hook(
...
+        reaped: summary.reaped,
+        plugins_matched: summary.plugins.len(),
+    }];
+
+    if session_start {
+        events.extend(
+            summary
+                .plugins
+                .into_iter()
+                .map(|pa| EventKind::PluginActivation {
```
nikomatsakis, line 376: same here this should be "plugin discovered" or something like that

nikomatsakis (changes requested):

gmemuriuki:
Going to close this. Was a fun exercise 😄

PR #262, opened 2026-07-13
RFD: predicate caching

kurasaiteja:
#### Updated V4
RFD for predicate caching. Custom predicates currently fork a process on every sync. This adds a small JSONL protocol that lets a predicate report the inputs its result depends on, so Symposium can cache the result and skip reevaluation until one of those inputs changes.

#### Protocol
We add three granular events on a #[non_exhaustive] enum:

- `WatchFile(PathBuf)` - result depends on this file
- `WatchEnv(String)` - result depends on this environment variable
- `WatchTime(usize)` - result becomes stale after this many milliseconds

A predicate can emit any number of events during one run. Its exit status still determines true or false. No events means cached indefinitely. `WatchTime(0)` means never cached. `shell(cmd)` emits WatchTime(0).

#### Storage
Cache lives at `~/.symposium/cache/predicates.json` and is discarded on Symposium version upgrade.

<details> <summary>Disclosure questions</summary>
AI disclosure.

I used an AI tool for research, autocomplete, or in other minimal ways. 
AI tool authored most of this RFD
</details>

### DIFF
--- a/md/rfds/predicate-caching/README.md
+++ b/md/rfds/predicate-caching/README.md
@@ -2,7 +2,7 @@
 
 ## TL;DR
 
-Predicates, especially custom predicates, spawn processes on every sync. This RFD lets custom predicates emit `Watch` JSONL events for files and environment variables. Symposium caches their results and skips reevaluation while the union of watched inputs is unchanged. No watch hints means cached indefinitely; `Volatile` means never cached.
+Predicates, especially custom predicates, spawn processes on every sync. This RFD lets custom predicates emit granular JSONL watch events for files, environment variables, and time. Symposium caches their results and skips reevaluation while every watched input is unchanged and no watch expires. No watch events means cached indefinitely; `WatchTime(0)` means never cached.
 
 ## Problem
 
@@ -10,35 +10,33 @@ Auto-sync means predicates re-evaluate on every agent session start. A workspace
 
 ## Design
 
-Custom predicates already emit JSONL events to stdout. We add two variants:
+Custom predicates already emit JSONL events to stdout. We add one event per watched resource:
 
 ```rust
-#[derive(Serialize, Deserialize)]
+#[non_exhaustive]
 enum CustomPredicateEvent {
     // ... existing variants ...
-    Watch {
-        files: Vec<PathBuf>,
-        env: HashMap<String, String>,
-    },
-    Volatile {},
+    /// Result depends on the contents of the given file.
+    WatchFile(PathBuf),
+    /// Result depends on the value of the given environment variable.
+    WatchEnv(String),
+    /// Result becomes stale after this many milliseconds.
+    WatchTime(usize),
 }
 ```
 
-A predicate can emit multiple watch hints:
+A predicate can emit any number of these events:
 
 ```jsonl
-{"watch": {"files": ["CargoBrazil.toml"]}}
-{"watch": {"files": ["Config"]}}
-{"watch": {"env": {"LAMBDA_ENV": "prod"}}}
+{"watchFile": "CargoBrazil.toml"}
+{"watchFile": "Config"}
+{"watchEnv": "LAMBDA_ENV"}
+{"watchTime": 60000}
 ```
 
-Symposium unions these into one watch set. A change to any watched input causes one reevaluation. Files are relative to the workspace root.
+Symposium unions the file and environment events. A change to any watched input or expiry of the shortest `WatchTime` causes one reevaluation. Files are relative to the workspace root.
 
-The process exit status determines the predicate result. `Watch` and `Volatile` events only control caching. No `Watch` events means the result is cached indefinitely and predicates must report every changing input or Symposium may reuse a stale result. `Volatile` disables caching and takes precedence over watch hints.
-
-```jsonl
-{"volatile": {}}
-```
+The process exit status determines the predicate result. Watch events only control caching. No watch events means the result is cached indefinitely; predicates must report every changing input or Symposium may reuse a stale result. `WatchTime(0)` means the result is stale immediately, which effectively disables caching. The `#[non_exhaustive]` attribute leaves room for new watch kinds.
 
 ## SDK helper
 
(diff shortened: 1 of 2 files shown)


md/rfds/predicate-caching/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Problem
+
+Auto-sync means predicates re-evaluate on every agent session start. A workspace with 10 plugins, each with a `shell(...)` gate, forks 10+ processes every time.
+
+## Design
+
+Two ways to declare watch paths:
+
+### 1. Syntactic
```
nikomatsakis: Hmm. I'm inclined to skip this and *just* focus on the runtime version. I guess that would make something like a `shell` predicate kind of inherently tricky but that's ok, I don't think an arbitrary shell command is a very good idea for portability reasons anyhow. I'd rather we have declarative forms for common shell commands.

md/rfds/predicate-caching/README.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+`watch` is a sibling field, not embedded in the predicate string.
+
+### 2. Runtime
+
+The shell command emits JSON to stdout:
+
+```json
+{"result": true, "watch": ["CargoBrazil.toml"]}
```
nikomatsakis: I think we restructured the custom predicates to emit lines of jsonl, right? I think that the idea would be that they emit a line like

```
{"watch": {
    "files": ["CargoBrazil.toml"]
}}
```

which would derive from

```rust
#[derive(SErialize, Deserialize)]
enum CustomPredicateEvent {
    ...., // whatever other events we have, I think right now that is for selected crates? I think we'll be removing that eventually but it's ok
    Watch { files: Vec<PathBuf> }
}
```

I believe that the predicate *result* comes from the exit status of the shell? Anyway, it should be independent from the custom predicate event.

md/rfds/predicate-caching/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+Commands that print nothing or non-JSON fall back to exit-code semantics with no caching. Backward compatible.
+
+Runtime takes priority when both are present. Either is sufficient to enable caching.
+
+## How it works
+
+Fingerprint is `mtime + size` (same model as Cargo). A missing file is a valid fingerprint state — invalidates when the file appears.
+
+Cache lives at `predicate-cache.json`, keyed by the normalized predicate string. Two plugins with the same predicate share one entry.
```
nikomatsakis: I think the key part is that the cache is stored in the user's symposium directory? I would assume we are kind of generalizing the current `Cargo.lock` caching mechanism.

md/rfds/predicate-caching/README.md
```diff
@@ -0,0 +1,81 @@
...
+4. On write, prune entries for predicates no longer in any manifest
+
+Cache is discarded on Symposium version upgrade. `watch = []` (empty) means static for the current invocation only — no persistent caching without at least one file to stat.
+
+## Built-in predicates
+
+- `workspace-member()` → cached, watch = [] (static per run)
+- `path_exists(path)` → cached, watches the path itself
+- `depends-on(name)` → cached, watches PM manifest (e.g. `Cargo.lock`)
+- `env(...)` → never cached (env isn't file-observable)
```
nikomatsakis: I think we should add another kind of predicate here - not just files but also the values of environment variables. Env can cache this way and shell predicates should also be able to emit that. We could have a helper in the `symposium-sdk` crate for reading from the environment that automatically emits the cache line too, so that when you read an env variable you are always cached on its value.

nikomatsakis (changes requested):
Thanks! Left some notes here.

md/rfds/predicate-caching/README.md
kurasaiteja: I'll drop the syntactic watch field and focus on the JSONL runtime approach. Agreed that arbitrary shell commands aren't great for portability and maybe declarative forms for common patterns is the better direction.

md/rfds/predicate-caching/README.md
kurasaiteja: I'll align with the existing `CustomPredicateEvent` enum and make 
```
Watch { files: Vec<PathBuf> } 
```
a variant. Exit status stays as the result, watch event is just a cache hint emitted alongside. Will update the RFD.

md/rfds/predicate-caching/README.md
kurasaiteja: Got it, I'll put it in `~/.symposium/cache/`. Since watch paths are resolved relative to the workspace root at stat time, the fingerprint comparison naturally should handle multi-workspace.

md/rfds/predicate-caching/README.md
```diff
@@ -1,62 +1,72 @@
...
-[[skills]]
-predicates = ["shell(grep -q lambda CargoBrazil.toml)"]
-watch = ["CargoBrazil.toml"]
-source.path = "skills/lambda"
+```jsonl
+{"watch": {"files": ["CargoBrazil.toml"], "env": {"LAMBDA_ENV": "prod"}}}
 ```
 
-`watch` is a sibling field, not embedded in the predicate string.
+The predicate result still comes from exit status — the watch event is just a cache hint. Predicates that don't emit a `Watch` event are never cached (current behavior preserved).
```
nikomatsakis: I think it should be the opposite. If you don't emit watch hint(s), you are ALWAYS cached, because we assume we don't have to cache.

Also, let's make it clear that predicates can emit multiple watch hints that are all unioned together.

md/rfds/predicate-caching/README.md
```diff
@@ -65,17 +75,16 @@ Cache is discarded on Symposium version upgrade. `watch = []` (empty) means stat
 ```rust
 struct DepsResult {
     ids: Vec<PackageId>,
-    watch: Option<Vec<WatchEntry>>,
+    watch: Option<Vec<PathBuf>>,
```
nikomatsakis: I don't think these should be optional. Also, do we need environments here?

I'm not exactly clear on the role of `DepsResult`, I guess I have to go digging.

nikomatsakis (changes requested):
Much better! Left a few more comments.

md/rfds/predicate-caching/README.md
kurasaiteja: I originally thought to extend `list_deps`(What packages does this workspace depend on?) to also return watch paths (`DepsResult { ids, watch }`) so that Symposium could cache the entire dependency list and skip re-calling the PM when manifest files haven't changed.

We can defer this because the PackageManager trait is still being defined in Jack's latest PR and the shape isn't settled yet. Also, the cargo PM's list_deps is already cheap (reads in-memory workspace data). Once PM interface stabilizes, we can add this.

md/rfds/predicate-caching/README.md
kurasaiteja: If a predicate author writes a shell command that depends on something but forgets to emit a watch event you mean the result gets cached forever?

nikomatsakis: Yes, I consider that user error. That said, we should probably add another kind of "watch" that is like "volatile", meaning: don't ever cache this. I would make the `shell` predicate itself "volatile".

md/rfds/predicate-caching/README.md
nikomatsakis: Yeah, I think we can defer it. This should focus on custom predicates which I don't *think* are involved in that path?

md/rfds/predicate-caching/README.md
```diff
@@ -2,89 +2,84 @@
...
 ## Problem
 
 Auto-sync means predicates re-evaluate on every agent session start. A workspace with 10 plugins, each with a custom predicate, forks 10+ processes every time.
 
 ## Design
 
-Custom predicates already emit JSONL events to stdout. We add a `Watch` event variant:
+Custom predicates already emit JSONL events to stdout. We add two variants:
 
 ```rust
```
nikomatsakis, line 15: Two things. First, I think we should make this more granular. Second, instead of `Volatile`, let's make it have a *timer*, which seems generally useful:

```rust
#[non_exhaustive]
enum CustomPredicateEvent {
    /// Value depends on the value of the given file 
    WatchFile(PathBuf),

    /// Value depends on the value of the given environment variable 
    WatchEnv(String),

    /// Value becomes stale in this number of milliseconds
    WatchTime(usize),
}
```

...this way, `WatchTime(0)` is effectively volatile. It also makes it easy to add new kinds of WatchEvents in the future.

nikomatsakis (changes requested):

nikomatsakis (approved):
Looks great!

md/rfds/predicate-caching/README.md
```diff
@@ -0,0 +1,83 @@
...
+{"watchTime": 60000}
+```
+
+Symposium unions the file and environment events. A change to any watched input or expiry of the shortest `WatchTime` causes one reevaluation. Files are relative to the workspace root.
+
+The process exit status determines the predicate result. Watch events only control caching. No watch events means the result is cached indefinitely; predicates must report every changing input or Symposium may reuse a stale result. `WatchTime(0)` means the result is stale immediately, which effectively disables caching. The `#[non_exhaustive]` attribute leaves room for new watch kinds.
+
+## SDK helper
+
+The `symposium-sdk` crate provides a helper that reads an environment variable and emits its watch event:
```
nikomatsakis, line 43: (we should probably also have a helper to e.g. read and deserialize a JSON/TOML file or just to read a file to a string)

nikomatsakis (commented):
Looks great!

PR #276, opened 2026-08-14
Feat: Remote MCP transports

Fluzko:
depends on #269 

## What does this PR do?

Until now Symposium could only talk to MCP servers it launched itself as child processes. A growing share of servers are hosted instead, reached over HTTP, and most of those want a real sign-in. This adds both: remote servers, and the authorization flow they need.

### For plugin authors

An `mcp_servers` entry with a `url` instead of a `command` is a remote server, reached over streamable HTTP:

```toml
[[mcp_servers]]
name = "my-server"
url = "https://mcp.example.com/mcp"
headers = { Authorization = "Bearer ${MY_TOKEN}" }
```

`url` and `headers` expand `${VAR}` and `${VAR:-default}`, so a manifest can name a credential without carrying its value. A variable that is unset and has no default refuses the server by name, rather than sending the literal `${VAR}` text and turning it into a confusing authentication failure.

`transport = "sse"` is rejected with a pointer to streamable HTTP. [The HTTP+SSE transport has been deprecated](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http#http+sse-transport-2024-11-05) since protocol revision 2025-03-26 and is eligible for removal.

### Signing in

Servers that require authorization are handled with OAuth, no token pasting:

```
cargo agents mcp login my-server     # opens the browser, stores the result
cargo agents mcp logout my-server
```

Tokens live one file per server under `~/.symposium/credentials`, and refresh on their own. Revoking one server is deleting one file.

Login is a command rather than something the meta-server does, because that process is a child of the agent: its stdout carries JSON-RPC and it has no terminal, so it can neither open a browser nor print a URL. When a server does need signing in, the meta-server says so by name and tells the agent the exact
command to suggest, and `cargo agents status` lists each remote server with whether a token is stored.

### Endpoint safety

A manifest names the destination, so the destination is checked. Plain HTTP is allowed only for loopback, which is already as trusted as the process itself. Private and link-local addresses are refused, both from the URL and again from every address it actually resolves to, since a public-looking hostname can answer with a private one. Redirects _are not_ followed.

### Known gap

Servers that do not offer dynamic client registration cannot be used yet, because _there is no way yet_ to supply a pre-registered client id. GitHub's MCP server is the notable example. TBD to discuss.

<details>

**AI disclosure.**

**Questions for reviewers.**

</details>


jackh726:
I'd like Niko to look over the meta MCP server PRs - but this seems orthogonal? Can you split out to a separate PR?

Fluzko:
> I'd like Niko to look over the meta MCP server PRs - but this seems orthogonal? Can you split out to a separate PR?

@jackh726 So this is not orthogonal, it's a direct extension over the work at #269. 
The thing is that I need that work to do this extension, but can't rebase nor point to that branch to see the actual diff over #269 because I don't have repo permissions.

If you'd like to review in the meantime, you cant start from https://github.com/symposium-dev/symposium/pull/276/commits/c8493f7f64603c7a63d644d5828a920cd10fc934 which is the first commit here.

PR #277, opened 2026-08-20
RFD: Agent Plugins interoperability

Fluzko:
## What does this PR do?

Proposes compiling Symposium plugins into the shared **Agent Plugins** directory format (manifest + `skills/`) that five of the seven supported agents already read, instead of writing skill files into a different per-agent path for each one.

- Compiles an already-gated plugin into one directory; agents that accept a registered path get pointed at it, others get a copy
- Reads an externally authored `plugin.json` package (crate, registry entry, or workspace member) as an ordinary Symposium plugin
- Adds Claude Code and Gemini CLI emitters over the same compiled model, since their own formats differ mainly in the manifest filename
- Installs through the existing `cargo agents sync` / `use` / `remove` path (no new command)
- Scopes MCP servers out of version one; they land with the [MCP meta-server](md/rfds/mcp-meta-server/README.md) design

Includes a follow-up commit addressing feedback left on the fork commit: states when directory removal actually happens (next sync where a plugin stops applying, not only on uninstall), forward-references the per-agent dialect breakdown instead of leaving it dangling, and defines what counts as a "broken" skill.

<details>

**AI disclosure.**

**Questions for reviewers.**

Does the version-one boundary (skills only, MCP servers deferred to the meta-server RFD) match the sequencing you'd expect? And does the "Claude Code / Gemini CLI get a separate emitter, same compiled model" framing hold up given how those two formats actually diverge from the standard?

</details>


md/rfds/agent-plugins/proposed-install.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Agents without a plugin unit
+
+OpenCode and Goose have no plugin directory to install into. OpenCode extends through TypeScript modules, and Goose through MCP servers. For these two, symposium installs skills the old way, into `.agents/skills/<skill-name>/`.
+
+## When two plugins want the same name
+
+The plain plugin name is used when it is free. When two plugins from different sources want the same name, or when a directory you manage yourself already sits there, symposium adds a short hash of the plugin's origin to the directory name instead.
+
+Cleanup keys on the `.symposium` marker rather than the directory name, so a plugin that moves between the two forms sorts itself out on the next sync.
```
gmemuriuki: "keys on" meaning:
1. candidate discovery, or
2. package association, or
3. deletion authority.

I'm not sure whether a `.symposium` marker is sufficient to authorize recursive deletion.

md/rfds/agent-plugins/README.md
```diff
@@ -0,0 +1,234 @@
+<!-- See .agents/skills/authoring-rfds/SKILL.md for style guidance -->
```
jackh726, line 1: Why is this here?

md/rfds/agent-plugins/README.md
```diff
@@ -0,0 +1,234 @@
...
+- Write each directory once and register its path with an agent where the agent accepts one, copying only where it does not.
+- Install through `cargo agents sync`, at the same scope as the enablement that selected the plugin. No new command.
+- Cover the format's skills component in version one. MCP servers are a follow-up that belongs to the meta-server design.
+
+## Motivation
+
+Symposium installs skills with one mechanism per agent. A skill directory goes to `.claude/skills/` for one agent, `.agents/skills/` for five, and `.kiro/skills/` for another, each with a separate global path. Symposium carries that knowledge itself, and supporting a new agent means learning another placement rule.
+
+The agents have since converged on a common answer. Every one of them that grew a plugin system settled on the same unit: a directory containing a manifest and a `skills/` folder. Three of the agents Symposium supports read the [Agent Plugins](https://agent-plugins.org/) format directly. Claude Code and Gemini CLI read formats of their own that differ mainly in the manifest filename. Producing that directory lets each agent's own loader perform the placement Symposium performs by hand.
```
jackh726, line 18: >The agents have since converged on a common answer.

You say this below, but this is not "technically" true: Claude, Gemini, Goose, and OpenCode haven't officially adopted Agent Plugins from what I can tell.

I think it'd be better to be clear that this is *one* (preferred) way that we expect to register components, but potentially not the only.

md/rfds/agent-plugins/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+### Agent backends
+
+Each backend renders the same compiled model and installs the result by whichever mechanism its agent supports:
+
+- GitHub Copilot and VS Code read the Agent Plugins format. The Copilot CLI stores packages under `~/.copilot/installed-plugins/`, and VS Code registers local plugin paths through a plugin-locations setting.
+- Codex CLI reads the format as of v0.147.0, from `.codex/plugins/` for a project and `~/.codex/plugins/` for a user.
+- Kiro reads the format as of v1.0.288, where a package is called a power and can be installed from a local folder.
+- Claude Code reads its own near-identical format, a `.claude-plugin/plugin.json` beside `skills/`, `agents/`, and `hooks/`, installed through a local directory marketplace. A package installed this way is recorded as installed but not enabled, so the enablement entry is written as well.
+- Gemini CLI reads extensions, a `gemini-extension.json` beside `skills/`, `commands/`, `hooks/`, and `agents/`, from `~/.gemini/extensions/<name>/`.
```
jackh726, line 148: I'm not *entirely* sure that this level of detail is needed in this RFD

jackh726 (approved):
couple small things, but LGTM overall

md/rfds/agent-plugins/README.md
Fluzko: leftover of the template, removed

md/rfds/agent-plugins/README.md
Fluzko: Fair, overstated.
 Rewritten so the convergence is on the unit, like:
- Codex CLI and the Copilot CLI read Agent Plugins, 
- Claude Code and Gemini/Antigravity CLI read their own format (close to Agent Plugins, portable) 
- OpenCode and Goose have no plugin unit. Preferred where an agent reads it, not the only mechanism, and that now sits in the TL;DR.

md/rfds/agent-plugins/README.md
Fluzko: Agreed, and it was partly wrong too. 
By implementing at #280, showed Codex CLI has no folder discovery, VS Code's `chat.pluginLocations` is machine-scoped, one directory can carry all three manifests so Claude Code and Gemini CLI need no separate emitters.
Paths and versions are moved to `proposed-install.md`, which already had the table. The README keeps just the invariant.

jackh726:
Can you rebase instead, instead of including merge commits?

PR #278, opened 2026-08-20
RFD for symposium uninstall

gmemuriuki:
## What does this PR do?
Proposes `cargo agents uninstall` for removing Symposium-managed integrations before `cargo uninstall symposium`.

Design covers:
1. Receipt backed backend
2. Project permits and global retirement tombstones
3. Missing binary hooks guards
4. Bounded cleanup without filesystem scan
5. Dry-run, blockers, shared, acknowledgements, retries, and recovery
6. User, tracked, shared, and ambiguous state preservations

RFD split into focused ownership, hook activation, cleanup engine, command references, and managed integrations chapters.

<details>

**AI disclosure.**

**Questions for reviewers.**

</details>

### DIFF
--- a/md/rfds/symposium-uninstall/README.md
+++ b/md/rfds/symposium-uninstall/README.md
@@ -2,35 +2,42 @@
 
 ## TL;DR
 
-- Add `cargo agents uninstall` to remove Symposium-managed integrations before `cargo uninstall symposium` removes the package.
-- Ask users to quit running agents before cleanup and restart them after Cargo removes the binary.
-- Record managed writes in durable receipts, then verify current identity before deleting anything.
-- Make generated hooks harmless when `cargo-agents` is absent and coordinate hook retirement during cleanup.
-- Preserve user-authored, shared, tracked, and ambiguous state by default.
-- Report what was removed or preserved and qualify the final removal assessment.
+- Add `cargo agents uninstall` to remove Symposium-managed integrations before Cargo removes the package.
+- Use receipts for bounded discovery and current artifact identity for deletion authority.
+- Make new hooks harmless when `cargo-agents` is absent and retire integrations one at a time.
+- Preserve user-authored, tracked, shared, modified, and ambiguous state by default.
+- Report what is clean in known scopes without claiming knowledge of unknown pre-receipt projects.
 
 ## Motivation
 
-`cargo uninstall symposium` removes the package binaries Cargo owns. It does not know that Symposium also wrote agent hooks, MCP entries, generated skills or plugin packages, caches, and workspace state.
+Today, `cargo uninstall symposium` removes the binaries Cargo installed. It does not remove hooks, MCP entries, generated skills, caches, or workspace state written by Symposium.
 
-An unguarded hook can therefore survive the binary:
+An unguarded global hook can therefore survive the binary and fail in every project:
 
 ```text
 /usr/bin/bash: line 1: cargo-agents: command not found
 PostToolUse:Bash hook error
 ```
 
-A stale global hook produces this error in every project. Project-scoped cleanup has a different problem: Symposium cannot scan arbitrary disks to rediscover every workspace, and it must not delete a similar-looking entry that another program or the user owns.
+Current init code can unregister known hook forms when an agent is removed from Symposium's configuration. It cannot inventory every managed side effect, prove who owns the current entry, or rediscover every project after the package is gone.
 
-The command needs bounded discovery, current ownership proof, safe hook retirement, and recovery from interrupted cleanup.
+Three cases define the required behavior:
 
-## As a user
+| Case | Required property |
+| --- | --- |
+| A global hook survives the binary | An absent binary is a quiet success |
+| The user or another program changes an entry | Preserve the current entry |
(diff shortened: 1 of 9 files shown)


md/rfds/symposium-uninstall/hook-activation/README.md
```diff
@@ -0,0 +1,173 @@
...
+POSIX machine-local global:
+
+```sh
+if [ -x <ABSOLUTE_PATH_POSIX_LITERAL> ]; then exec <ABSOLUTE_PATH_POSIX_LITERAL> hook <AGENT> <EVENT> --managed-id <UUID>; elif command -v cargo-agents >/dev/null 2>&1; then exec cargo-agents hook <AGENT> <EVENT> --managed-id <UUID>; else exit 0; fi
+```
+
+POSIX portable project:
+
+```sh
+if [ -x ${CARGO_HOME:-$HOME/.cargo}/bin/cargo-agents ]; then exec ${CARGO_HOME:-$HOME/.cargo}/bin/cargo-agents hook <AGENT> <EVENT> --managed-id <UUID>; elif command -v cargo-agents >/dev/null 2>&1; then exec cargo-agents hook <AGENT> <EVENT> --managed-id <UUID>; else exit 0; fi
```
Fluzko, line 122: careful here, a path with spaces will break, say the username is "John Doe" instead of a more classical "johndoe"

md/rfds/symposium-uninstall/README.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Why must agents be quit first?
+
+An agent may keep settings in memory and rewrite an old file on exit. Quitting first removes that race. Restarting after package removal reloads the cleaned configuration.
+
+## Implementation plan and status
+
+Implementation has not begun. The steps are dependency-ordered.
+
+### Step 1: Establish ownership primitives
```
Fluzko, line 216: This step won't ship anything and seems huge, so feels a bit untestable. What about shipping an e2e solution for one specific agent, say Claude?

md/rfds/symposium-uninstall/cleanup-engine/README.md
```diff
@@ -0,0 +1,186 @@
...
+Before planning a project-file mutation, uninstall checks ancestors for a `.git` file or directory without launching Git. If none exists, Git is not needed.
+
+When a repository exists:
+
+- tracked configuration is preserved as project-committed by default;
+- `--include-tracked` authorizes removal only of proven Symposium structure;
+- unavailable or indeterminate Git tracking preserves the file as a blocker; and
+- reports identify the file and structural locator without secrets.
+
+Acknowledgement may transfer an entry to the user. It cannot make a live unguarded `cargo-agents` hook or MCP invocation safe for package removal. That reference must be removed manually or with `--include-tracked`.
```
Fluzko, line 75: Hooks are saved as a shell command, so the guard fits around them. MCP entries are saved as `command` plus `args` and run directly, so there is no place for a guard. After `cargo uninstall symposium`, every stale MCP entry fails to start in every agent, which is the error the Motivation section opens with.

The plan is "`run cargo agents uninstall first`", but Motivation says people go straight to `cargo uninstall symposium today`. Once that's done the cleanup tool is gone and it's not clear to me what to do next.

Can we get the recovery steps written down, and a line on why MCP entries do not get a guard? Also I feel worth moving into Drawbacks, since Scope lists hooks and MCP together and reads like both are covered.

md/rfds/symposium-uninstall/cargo-agents-uninstall/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+If no ancestor contains a `.git` file or directory, uninstall treats the project as outside Git and does not require a Git executable.
+
+## Acknowledging a blocker
+
+`--acknowledge <BLOCKER-ID>` preserves an artifact and transfers responsibility to you. The report gives its location and a redacted manual edit. This is not `--force` and never weakens ownership checks.
+
+The ID stays stable for the same artifact kind, adapter, normalized target, and structural location. Changing the artifact invalidates the acknowledgement.
+
+An unguarded hook or MCP server that still launches `cargo-agents` cannot be acknowledged into a ready result. Remove that live reference before uninstalling the package.
```
jackh726, line 93: I'm not sure this is super useful. I think it should just be `--force`.

md/rfds/symposium-uninstall/cleanup-engine/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+```text
+artifact type + adapter + normalized target + structural locator
+```
+
+The acknowledgement stores that locator and the artifact's current identity, not only its display ID. Moving the locator creates a new blocker; changing the artifact invalidates acknowledgement.
+
+Acknowledgement preserves the artifact, records user responsibility, retires Symposium's claim, prints a redacted manual edit, and makes later installation treat the occupied slot as a structural collision.
+
+Successful finalization may delete acknowledgements because collision detection inspects the occupied entry. There is no `--force`: bypassing identity checks could delete user or third-party state.
```
jackh726, line 141: I think we should just not do this.  If uninstall isn't possible "safely", then the user should pass `--force`. I don't think it makes a lot of sense to *not* have an "all or nothing"

PR #281, opened 2026-08-31
RFD: Replace Gemini CLI with Antigravity CLI

Fluzko:
## What does this PR do?

Adds an RFD proposing we drop the `gemini` agent and support Antigravity CLI (`agy`) instead.

Gemini CLI's consumer sign-in is closed — "Sign in with Google" now fails with a message from Google telling you to migrate to Antigravity. That's the path our users take, so supporting Gemini means supporting an agent most people can't log into.

Antigravity is not a rename. Skills and MCP map over cleanly (it reads the same `.agents/skills/` layout we already write, and the familiar `mcpServers` shape), but hooks are a new file, new shape, new event names, new payload fields and a new output contract. That's why this replaces the agent rather than adding one.

Two findings from testing `agy` 1.1.22 that shaped the design:

- A `PreToolUse` hook returning `{}` **denies the tool call**. Our dispatcher returns `{}` whenever no plugin contributed, which is the common case — so a naive port would block every tool call. Exit codes are ignored entirely.
- There's no `SessionStart` event. The RFD derives one from `PreInvocation` plus the first sighting of a `conversationId`, since `invocationNum` resets every turn.

The RFD includes a proposed agent-details reference page and a six-step implementation plan. No code changes here.

<details>

**AI disclosure.**

Claude did the research and drafted the RFD, including running an instrumented `agy` locally to verify the hook contract rather than relying on docs. Direction, scope and the design calls were mine.

**Questions for reviewers.**

- Is replacing Gemini the right call, or should both be supported for a transition period?
- The session-start substitute (tracking seen `conversationId`s) is the least conventional part of the design — worth a look.

</details>

### DIFF
--- a/md/SUMMARY.md
+++ b/md/SUMMARY.md
@@ -96,6 +96,8 @@
       - [Discovery & sync](./rfds/registry-centric-plugins/discovery-sync/README.md)
       - [User-managed plugins](./rfds/registry-centric-plugins/user-managed-plugins/README.md)
     - [Predicate caching](./rfds/predicate-caching/README.md)
+    - [Replace Gemini CLI with Antigravity CLI](./rfds/antigravity-cli/README.md)
+      - [Proposed agent details](./rfds/antigravity-cli/proposed-agent-details.md)
   - [Completed](./rfds/completed.md) <!-- move completed rfds to this section -->
     - [Configuration parsing and normalization](./rfds/config-normalization/README.md)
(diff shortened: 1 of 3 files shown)


jackh726 (commented):
Not sure we need this (I think of RFDs about design - and this is just about updating our support).

I can merge if you think we should; otherwise, let's just merge the impl PR

PR #282, opened 2026-09-01
Feat: Replace Gemini CLI with Antigravity CLI

Fluzko:
depends on #274 
needs some work over plugins when merged #280 

## What does this PR do?

Adds Google's Antigravity CLI (`agy`) as a supported agent and retires Gemini CLI. Implements #281.

Gemini CLI's consumer sign-in is closed, "Sign in with Google" now fails with a message from Google telling you to migrate to Antigravity. That's the path our users take, so we follow.

Three commits, reviewable in order:

1. Fix hook scope: sync registers hooks at the configured scope (see "One fix worth a look")
2. Replace the agent: Antigravity in, Gemini out, skills, MCP servers and hooks
3. Document it: reference page, agent-details page, and the usual tables

Plus three review follow-ups on top: merge `main`, move the prompt-event gate out of the dispatcher, and drop the compatibility shims.

### Antigravity in short

Skills were free: it reads the same `.agents/skills/` layout we already write for Copilot, Codex, OpenCode and Goose. MCP was easy: its own `mcp_config.json`, familiar `mcpServers` shape.

Hooks were the work. A few of its behaviours are unusual enough to shape the code, and each was verified by running `agy` rather than read from its docs:

- **A `PreToolUse` hook that returns `{}` denies the tool call**, and exit codes are ignored entirely. Our dispatcher returns `{}` whenever no plugin contributes, the common case, so without care this would block every tool call. We always emit an explicit allow.
- **`SessionStart` exists but isn't documented.** The docs list five events; the binary has six. It works, and fires with the first turn of a new conversation, never on resume.
- **There's no prompt event.** `PreInvocation` stands in, but fires before every model call, so we only act on the first of a turn. That rule lives on the payload type (`AgentHookInput::should_dispatch`), so the dispatcher stays agent-agnostic.

### Retiring Gemini

No compatibility shims, per review. After upgrading, a `gemini` entry left in `~/.symposium/config.toml` makes `cargo agents sync` fail with `unknown agent: gemini` (hooks keep running, without auto-sync) until you remove it or re-run `cargo agents init`; a hook registration left in `.gemini/settings.json` needs removing by hand; a plugin manifest declaring `format = "gemini"` is skipped with a load warning.

### One fix worth a look

`sync` registered hooks and MCP by passing the workspace root to the *global* functions. That only produces the right path when an agent's project and global locations happen to share a shape. Antigravity's don't, so project scope wrote to a file `agy` never reads.

The project-scoped functions already existed but had no callers. `sync` now uses them, **which also moves Copilot's project hooks from `.copilot/settings.json` to `.github/hooks/`, where Copilot actually reads them.** Copilot users on project scope get working hooks for the first time; a dead `.copilot/settings.json` stays behind in their repos.

### Testing

472 tests pass. Beyond unit and integration coverage, the agent was driven end to end against a real `agy` (1.2.7) to confirm hooks load, tool calls aren't denied, and session-start fires once per new conversation and not on resume.

<details>

**AI disclosure.**

Claude did the research, implementation and testing, including probing a real `agy` install to establish the hook contract. Direction, scope and the design calls were mine.

**Questions for reviewers.**

- The Copilot path change is a fix, but it's a behaviour change riding along in a PR about a different agent. Happy to split it out if you'd rather.
- MCP registration is covered by unit tests but hasn't been exercised against a live `agy`, that needs a plugin declaring an MCP server, which depends on other in-flight work.

</details>


src/bin/cargo-agents.rs
```diff
@@ -141,7 +145,13 @@ async fn main() -> ExitCode {
...
     match cli.command {
         // Commands that need direct I/O (stdin/stdout) stay in the binary
-        Some(Commands::Hook { agent, event }) => hook::run(&sym, agent, event).await,
+        // A retired agent parses to `None`: a hook registration left behind by
+        // an agent symposium no longer supports exits cleanly instead of failing
+        // in that agent's session on every event.
+        Some(Commands::Hook { agent, event }) => match agent.0 {
+            Some(agent) => hook::run(&sym, agent, event).await,
+            None => ExitCode::SUCCESS,
+        },
```
jackh726, line 154: This feels...not great.

src/hook.rs
```diff
@@ -235,6 +235,18 @@ pub async fn execute_hook(
...
+    // user-prompt hooks would run several times in a single turn. Empty stdout
+    // is Antigravity's "no opinion", so returning nothing is safe.
+    if agent == HookAgent::Antigravity
+        && event == HookEvent::UserPromptSubmit
+        && !crate::hook_schema::antigravity::is_first_invocation(input)
+    {
+        tracing::debug!("antigravity: not the first invocation of the turn, skipping prompt event");
+        return Ok(Vec::new());
+    }
+
```
jackh726, line 249: I don't think we should have agent-specific code here. If there is some generalized mechanism we need to add, then we should do that.

I'm also not quite sure what this actually is for? *Why* does antigravity fire multiple of these? Likely it's because *every* prompt fires, right?

Or, is it the case that a single "prompt" gets multiple PreInvocation calls?

src/init.rs
```diff
@@ -31,28 +31,40 @@ fn resolve_agents(
...
+            if !Agent::is_retired(name) {
+                Agent::from_config_name(name)?;
+            }
             names.retain(|n| n != name);
         }
-        return names.iter().map(|n| Agent::from_config_name(n)).collect();
+        return names
+            .iter()
+            .filter_map(|n| Agent::from_configured_name(n, out).transpose())
+            .collect();
```
jackh726, line 59: What is the difference between `Agent::from_configured_name` and `Agent::from_config_name`? This changed?

jackh726: I guess one returns an Option probably? Seems like one should be `from_config_name` and the other should be `opt_from_config_name`.

src/migrations.rs
```diff
@@ -0,0 +1,176 @@
+//! One-shot cleanups for state a past release wrote and the current one no
```
jackh726, line 1: I'm pretty tempted to say that we shouldn't do this. I think barely anyone uses Symposium right now, and even fewer (if anyone) uses Symposium + Gemini.

Similar with all the "retired" things in this PR. I think we should be very careful to not add too much "backwards-compatibility" here, because I just don't think it's needed and I'd rather keep the code clean(er).

src/plugins.rs
```diff
@@ -832,25 +832,40 @@ pub enum HookFormat {
...
     Copilot,
-    Gemini,
     Kiro,
+    /// A wire format symposium has retired along with its agent.
+    ///
+    /// A published manifest outlives the release that drops an agent, so the
+    /// spelling still deserializes; the hook is skipped rather than failing the
+    /// plugin that carries it.
+    #[serde(rename = "gemini")]
+    Retired,
```
jackh726, line 846: Can you do multiple `#[serde(rename = ...)]`? If not, then this doesn't scale to if we retire other agents.

src/state.rs
```diff
@@ -25,20 +25,30 @@ pub struct State {
     /// Last time we checked crates.io for a newer version.
     #[serde(default, rename = "last-update-check")]
     pub last_update_check: Option<DateTime<Utc>>,
+
+    /// Ids of one-shot migrations already applied to this directory.
+    ///
+    /// Recorded by id rather than inferred from `version` so a migration runs
+    /// exactly once regardless of which release the user upgraded from.
+    #[serde(default, rename = "applied-migrations")]
+    pub applied_migrations: Vec<String>,
```
jackh726, line 34: See my comment before: this feels a bit cart-before-the-horse, and I don't think we should bother.

jackh726 (commented):
Sorry for taking a while to review this. The biggest thought here is that we should just make this change without worrying about migrations or retiring or anything, since I don't think it is going to matter in practice.

src/bin/cargo-agents.rs
Fluzko: Agreed, gone. `HookAgentArg` is removed and the arm is back to `hook::run(&sym, agent, event)`. A leftover `cargo-agents hook gemini ...` registration now fails clap parsing like any unknown agent.

src/init.rs
Fluzko: yes, the difference was that it returned `Option`, warns and skips for a retired name. Both gone now. 
Only `from_config_name` remains, unchanged from before this PR.

src/migrations.rs
Fluzko: Agreed, removed. 
Also removed `RETIRED_AGENTS`, `HookFormat::Retired`, `HookAgentArg` and `applied-migrations`. A stale Gemini entry now makes `cargo agents sync` fail with `unknown agent: gemini (expected antigravity, ...)`. Hooks keep running but skip auto-sync (warning in the log) until the entry is removed, by hand or by re-running `cargo agents init`. 
Noted also in the PR description.

src/plugins.rs
Fluzko: `#[serde(alias = "...")]` is repeatable, so it would have scaled. But the variant is gone: `format = "gemini"` now fails to deserialize and the plugin is skipped with the usual load warning, `unknown variant 'gemini', expected one of 'symposium', 'antigravity', ...`

src/hook.rs
Fluzko: Both strictly, because `PreInvocation` fires before every model call, and one prompt makes several model calls whenever the model uses tools: one per tool round plus the final answer. Probed agy 1.2.7 with a logging hook; one prompt that ran `ls` and then read a file produced:

```
PreInvocation invocationNum=0
PreToolUse / PostToolUse
PreInvocation invocationNum=1
PreToolUse / PostToolUse
PreInvocation invocationNum=2
Stop
```

The next prompt started again at 0. 
The docs say that invocationNum is "the 0-indexed sequence number of the current model invocation". And there is genuinely no prompt event to use instead: the binary's hook proto has exactly six payload types (`PreTool`, `PostTool`, `PreInvocation`, `PostInvocation`, `SessionStart`, `Stop`).

So I've replaced the agent-specific branch with a trait method, `AgentHookInput::should_dispatch`, default true. Antigravity's `PreInvocation` payload implements it as `invocationNum == 0`. When it returns false, dispatch answers with the agent's neutral output, the same bytes an uncontributed first invocation returns `({"injectSteps":[]})`, and runs nothing. `hook.rs` no longer names any agent. 
The seam is reusable: Claude, Codex and Copilot `SessionStart` carry `source = startup|resume|compact|clear`, and we currently re-run the discovery hint on all of them.

If you'd rather not map `PreInvocation` to user-prompt-submit at all, I'm fine dropping the mapping. Nothing builtin consumes it and Antigravity never sends the prompt text anyway.

Fluzko:
@jackh726  Done. Every compat shim and the migration are gone. The branch now adds Antigravity and removes Gemini, nothing in between. Merged main to clear the conflict. The Copilot project-hooks fix stays, since it's unrelated to retirement.
Let me know if you'd like another pass over this

jackh726:
Please rebase instead of merging.

PR #283, opened 2026-09-01
 Add a focused Criterion benchmark suite

gmemuriuki:
## What does this PR do?
This PR adds the first focused benchmark suite for Symposium.
Rather than starting with a broad collection of microbenchmarks, it answers one concrete 
performance question: how much latency does Symposium's in-process `PreToolUse` path adds to
unchanged workspace?

The initial suite measures four cases:
- workspace dependency resolution with an empty Symposium cache
- dependency resolution through a new disk resolver with a valid disk cache
- `PreToolUse` dispatch with minimal configuration
- `PreToolUse` dispatch with a representative local plugin registry.

Pairing the component and hook measurements makes the result easier interpret. In particular
it shows how much work the workspace cache avoids and makes the existing Cargo workspace-lookup
cost visible.

### what changed
 - Added a non-publishable `symposium-benchsuite` workspace package using Criterion.
 - Added deterministic Cargo workspace, configuration, and local registry fixtures.
 - Added reusable fixture staging, sandboxing, validation, and mock-Cargo support.
 - Added integration tests covering resolver memoization, disk-cache reuse, and lockfile invalidation.
 - Documented a measurement contract beside every benchmark target.
 - Added normal CI checks that compile benchmark targets on native Linux, macOS, and Windows jobs.
 - Added a dedicated benchmark workflow for manual runs and relevant pull requests.
 - Updated existing workflows to use Node 24-compatible GitHub Actions.

<details>

**AI disclosure.**

* Other: I used Codex and Claude throughout this work for research, design discussion, code review, and some implementation and editing. I reviewed and approved the resulting changes and made the final design decisions.

**Questions for reviewers.**

</details>


benches/benchsuite/src/cargo.rs
Fluzko, line 1: The #[cfg(windows)] batch shims never executes in CI: the Windows job only does cargo check, and the benchmarks workflow is Linux-only. 
The native jobs should run `cargo test -p symposium-benchsuite --lib` too? It's nearly cero time and would actually exercise the shim plus the fixture validation on macOS and Windows

md/design/benchmarking.md
```diff
@@ -0,0 +1,273 @@
...
+
+The wrapper is never part of a timed sample.
+
+Criterion targets support a fast smoke run through:
+
+```text
+cargo test -p symposium-benchsuite --benches
+```
+
+Smoke runs execute workloads without collecting full measurements. Normal pull request CI compiles benchmark targets and runs the small support-library and cache-invariant tests. Full smoke and measurement runs belong to the benchmark workflow.
```
Fluzko, line 211: I think that support library tests are not ran. They only run inside the path-filtered Benchmarks workflow (where `--benches` happens to also pull in the lib unittests).
Also says that `cargo package --list` "verifies" that benches are excluded, but nothing in `xtask/` or `.github/` runs it.

PR #286, opened 2026-09-11
Build the telemetry recording foundation

gmemuriuki:
## What does this PR do?

This builds the foundation for the accepted telemetry RFD (#273).
                                                                                                                                                                                                               
It adds strict, versioned schemas for the first telemetry rows, typed classification of stored lines, scoped HMAC identifiers, and private in-memory identity state. Cohort, window-rollover, and reset transitions are defined without adding persistence or production event emission.

It also defines the initial package-resolution vocabulary and clarifies that a future uploader will read validated records from closed days.

### Verification                                                                                                                                                                                             

- 140 telemetry tests pass.
- Clippy passes with warnings denied.
- The mdBook builds with two existing missing-example warnings.

<details>
                                                                                                                                                                                                               
**AI disclosure.**

- The AI tool authored large parts of the code.

**Questions for reviewers.**

- Does this provide the right boundary for the storage work that follows?

  </details>


src/telemetry/schema/agent.rs
```diff
@@ -0,0 +1,518 @@
...
+    pub(in crate::telemetry) start: SessionStartKind,
+    pub(in crate::telemetry) session_id: Option<SessionId>,
+    pub(in crate::telemetry) retention_subject: RetentionSubject,
+    pub(in crate::telemetry) cohort_day: CohortDay,
+}
+
+/// Version 1 record of a completed registered session-start hook.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(in crate::telemetry) struct SessionStartV1 {
```
kurasaiteja, line 149: Nit: `SessionStartV1::new` derives day from at, so rows written through the normal path are consistent. Deserialization does not enforce that relationship, so it can accept a row where day and at refer to different dates? Should we validate that here, or is the future archive reader expected to own this check?

gmemuriuki: Aah, yeah. I see. I'll fix that and validate it during deserialization and classify mismatch as invalid. 👍

gmemuriuki: Nice catch! thanks.

Fluzko:
This looks pretty nice, I've made my agents play around with this solution & the #287 one.
A few minor concerns that might be worth looking:

The identity API here gets reworked in #287, so thing like `IdentityDeriver`, `ScopedDimension` and `IdentityWindow` are gone there, and get replaced by other entities.  They're only ever exercised by #286's own tests in the meantime. Not asking to change it, but the diff would look nicer and easier to review both PRs in one shot. 

`ExtensionResolutionV1::new` takes a `path`, but there's no way to build one from outside When you'll write the producer you'll face this on day 1.

`telemetry-state.toml` it's locked to version 1 and rejects unknown fields, but the contract says v1 also carries the high-water mark, marker metadata and the session sets, and those aren't in the struct yet. When storage adds them it'll have to either call it v2 and migrate a format nobody ever had on disk, or make them optional. Fine while nothing has shipped, I'd just avoid cutting a release with this file format before storage lands.

jackh726:
Want to review this - but can reduce the number of commits here and in other PRs? I try to review commit-by-commit typically, and that's just impractical here.

gmemuriuki:
Doing that right now, give me 1 hour before you review ( If not yet started ) so I can make it better.

gmemuriuki:
I'm going to close this PR for it is superseded by another one which addresses the above concerns. The link to it is https://github.com/symposium-dev/symposium/pull/296

Sorry all for this mixup. I'll improve next time 👍

